PR #2, opened 2016-03-21
Login view layout

danielaRiesgo:
## Sumary ##
**Trello card**: https://trello.com/c/mHc2j1xc/4-vista-log-in

Added the default login view (xib file) and added to the logic the elements of the view that weren't binded yet.
Added a demo example of how to use the Authentication Boostrapper and be able to manually test the login process.

## Test Plan ##

(Don't click on "forgot password ?" and/or "Register"!)

  * Open the app
  * The only email-password combination that works is "example@mail.com" and "password".
  * Tryout validation errors and login errors.

### iPhone 4s
Initial screen
4s normal
Validation Errors with Keyboard
4s errores email
4s errores password
Login enabled
4s login enabled
Login executing
4s login ejecutandose
Bad login
4s login error
4s login fallido

### iPhone 5
Initial screen
5 normal
Validation Errors with Keyboard
5 errores email
5 errores password
Login enabled
5 login enabled
Login executing
5 login ejecutandose
Bad login
5 login error
5 login fallido

### iPhone 6
Initial screen
6 normal
Validation Errors with Keyboard
6 errores email
6 errores password
Login enabled
6 login enabled
Login executing
6 login ejecutandose
Bad login
6 login error
6 login fallido

### iPhone 6s+
Initial screen
6s normal
Validation Errors with Keyboard
6s errores email
6s errores password
Login enabled
6s login enabled
Login executing
6s login ejecutandose
Bad login
6s login error
6s login fallido

## Known issues ##

1. ~~Frame display when keyboard is showing looks good on iphone 6 and 6s plus but not in 4 and 5. (**Trello card**: https://trello.com/c/AlZ61WAX/21-arreglar-display-al-mostrar-el-teclado )~~
2. ~~In the password textfield and button, the textfield border is visible against the button although very dim. (**Trello card**: https://trello.com/c/MPUsMM3F/20-corregir-el-borde-del-textfield-para-la-contrasena-en-la-vista-de-login )~~
3. When the security option is set on the password text field and the following actions happen, the already written password erases and starts writting again:
    Write in password text field.
    Change to email textfield or click any other element OR touch show and hide again.
    Click again in password text field to continue writing.
(**Trello card**: https://trello.com/c/B3D4IKhw/22-bug-al-reescribir-contrasena-en-security-mode )
4. ~~When the user selects to show password, the font and size is not what expected. (**Trello card**: https://trello.com/c/WD4SbyMV/24-mostrar-contrasena-cambia-la-font-de-forma-no-deseada )~~
5. Strings files still have to be associated with the app manually. (**Trello card**: https://trello.com/c/bQyWymgh/23-lograr-string-localization-automatica )
6. ~~Colours can't be customized yet. (**Trello card**: https://trello.com/c/7WPRPJEH/19-configurar-la-paleta-de-colores-en-el-log-in )~~


guidomb:
@danielaRiesgo It seems that the passworld field has a double border, probably caused by the text field its self. Can you try to remove the borders from the text field?

guidomb:
@danielaRiesgo I've just realized this is a known issue. Can you create a card to fix this and put the card link in the PR?

guidomb:
Create a card for every know issue, and if you know how to fix it put in the card's description and then add each card's link in this PR with its corresponding known issue.

Authentication/LogIn/View/LoginController.swift
```diff
@@ -48,10 +55,12 @@ public final class LoginController: UIViewController {
     init(viewModel: LoginViewModelType,
         loginViewFactory: () -> LoginViewType,
         onRegister: (LoginController) -> (),
-        delegate: LoginControllerDelegateType = DefaultLoginControllerDelegate()) {
+        onRecoverPassword: (LoginController) -> (),
```
guidomb: I think that this initializer is getting too convoluted. Maybe we can move the `onRecoverPassword` and `onRegister` to its own delegate. The difference being we wont provide a default delegate. This must be provided by the user.

Authentication/LogIn/View/LoginController.swift
```diff
@@ -97,6 +117,13 @@ private extension LoginController {
         _viewModel.email <~ loginView.emailTextField.rex_textSignal
         loginView.emailLabel.text = _viewModel.emailText
         loginView.emailTextField.placeholder = _viewModel.emailPlaceholderText
+        _viewModel.emailValidationErrors.signal.observeNext { [unowned self] errors in
+            if errors == [] {
```
guidomb: I think its would be better to do `error.isEmpty`

Authentication/LogIn/View/LoginController.swift
```diff
@@ -16,18 +16,23 @@ import enum Result.NoError
...
     private let _onRegister: (LoginController) -> ()
+    private let _onRecoverPassword: (LoginController) -> ()
     private let _loginViewFactory: () -> LoginViewType
-    private let _delegate: LoginControllerDelegateType
+    private let _delegate: LoginControllerDelegate
 
     public lazy var loginView: LoginViewType = self._loginViewFactory()
     
+    private lazy var _notificationCenter: NSNotificationCenter = NSNotificationCenter.defaultCenter()
+    private var _notificationDisposables: [Disposable] = []
```
guidomb: This should be `private var _disposable = CompositeDisposable()`

Authentication/LogIn/View/LoginController.swift
```diff
@@ -125,12 +158,75 @@ private extension LoginController {
...
+        loginView.recoverPasswordButton.setTitle(_viewModel.recoverPasswordButtonTitle, forState: .Normal)
+        loginView.recoverPasswordButton.setAction { [unowned self] _ in self._onRecoverPassword(self) }
+    }
+    
+}
+
+extension LoginController {
+    
+    public func addKeyboardObservers() {
+        _notificationDisposables.append(_notificationCenter.rac_notifications(UIKeyboardDidHideNotification)
```
guidomb: I would prefer the following

```swift
_disposable += _keyboardDisplayed <~ _notificationCenter.rac_notifications(UIKeyboardDidHideNotification)
  .map { _ in false }
```

Authentication/LogIn/View/LoginController.swift
```diff
@@ -16,18 +16,23 @@ import enum Result.NoError
...
+    private let _onRecoverPassword: (LoginController) -> ()
     private let _loginViewFactory: () -> LoginViewType
-    private let _delegate: LoginControllerDelegateType
+    private let _delegate: LoginControllerDelegate
 
     public lazy var loginView: LoginViewType = self._loginViewFactory()
     
+    private lazy var _notificationCenter: NSNotificationCenter = NSNotificationCenter.defaultCenter()
+    private var _notificationDisposables: [Disposable] = []
+    public private(set) var keyboardDisplayed: Bool = false
```
guidomb: This should be `private let _keyboardDisplayed = MutableProperty(false)`

Authentication/LogIn/View/LoginController.swift
```diff
@@ -125,12 +158,75 @@ private extension LoginController {
...
+        }
+    }
+    
+    func keyboardWillShow(notification: NSNotification) {
+        if let keyboardSize = (notification.userInfo?[UIKeyboardFrameEndUserInfoKey] as? NSValue)?.CGRectValue() {
+            if !keyboardDisplayed {
+                keyboardDisplayed = true
+                let keyboardOffset = keyboardSize.height
+                let emailOffset = loginView.emailTextField.convertPoint(loginView.emailTextField.frame.origin, toView: self.view).y - 10
+                //print("Email offset: \(emailOffset)")
```
guidomb: Remove all commented print statements

Authentication/LogIn/View/LoginController.swift
```diff
@@ -125,12 +158,75 @@ private extension LoginController {
...
+    public func addKeyboardObservers() {
+        _notificationDisposables.append(_notificationCenter.rac_notifications(UIKeyboardDidHideNotification)
+            .startWithNext { [unowned self] in
+                self.keyboardDidHide($0)
+        })
+        _notificationDisposables.append(_notificationCenter.rac_notifications(UIKeyboardWillShowNotification)
+            .startWithNext { [unowned self] in
+                self.keyboardWillShow($0)
+        })
+        _notificationDisposables.append(_notificationCenter.rac_notifications(UIKeyboardWillHideNotification)
```
guidomb:
```swift
_disposable += _notificationCenter
  .rac_notifications(UIKeyboardWillHideNotification)
  .startWithNext { [unowned self] in self.view.frame.origin.y = 0 }
```

Authentication/LogIn/View/LoginController.swift
```diff
@@ -125,12 +158,75 @@ private extension LoginController {
...
+}
+
+extension LoginController {
+    
+    public func addKeyboardObservers() {
+        _notificationDisposables.append(_notificationCenter.rac_notifications(UIKeyboardDidHideNotification)
+            .startWithNext { [unowned self] in
+                self.keyboardDidHide($0)
+        })
+        _notificationDisposables.append(_notificationCenter.rac_notifications(UIKeyboardWillShowNotification)
```
guidomb:
```swift
_disposable += _notificationCenter
  .rac_notifications(UIKeyboardWillShowNotification)
  .startWithNext { [unowned self] in self.keyboardWillShow($0) }
```

Authentication/LogIn/View/LoginController.swift
```diff
@@ -125,12 +158,75 @@ private extension LoginController {
...
+            .startWithNext { [unowned self] in
+                self.keyboardWillShow($0)
+        })
+        _notificationDisposables.append(_notificationCenter.rac_notifications(UIKeyboardWillHideNotification)
+            .startWithNext { [unowned self] in
+                self.keyboardWillHide($0)
+        })
+    }
+    
+    public func removeKeyboardObservers() {
```
guidomb: Delete this method and then add 

```swift
deinit {
  _disposable.dispose()
}
```

Authentication/LogIn/View/LoginControllerDelegate.swift
```diff
@@ -22,21 +22,30 @@ public protocol LoginControllerDelegateType {
...
+    
 }
 
-extension LoginControllerDelegateType {
+extension LoginControllerDelegate {
     
     public func loginControllerWillExecuteLogIn(controller: LoginController) {
-        controller.loginView.activityIndicator.startAnimating()
-        UIApplication.sharedApplication().beginIgnoringInteractionEvents()
+        // Start activity indicator
```
guidomb: You should do this

Authentication/LogIn/View/LoginControllerDelegate.swift
```diff
@@ -22,21 +22,30 @@ public protocol LoginControllerDelegateType {
...
     }
     
     public func loginControllerDidExecuteLogIn(controller: LoginController) {
-        controller.loginView.activityIndicator.stopAnimating()
-        UIApplication.sharedApplication().endIgnoringInteractionEvents()
+        // Finish activity indicator
```
guidomb: Same here

Authentication/RecoverPassword/RecoverPasswordController.swift
```diff
@@ -0,0 +1,14 @@
...
+//  Authentication
+//
+//  Created by Daniela Riesgo on 3/18/16.
+//  Copyright © 2016 Wolox. All rights reserved.
+//
+
+import Foundation
+
+
+public final class RecoverPasswordController: UIViewController {
```
guidomb: This shouldn't be part of this PR it is out of the scope of the card. You can leave it but next time only do what is necessary for this feature to work.

AuthenticationDemo/ExampleSessionService.swift
```diff
@@ -0,0 +1,67 @@
...
+    
+    init(email: String, password: String) {
+        _email = email
+        _password = password
+        currentUser = AnyProperty(initialValue: Optional.None, signal: _currentUser.map { $0 })
+        (events, _eventsObserver) = Signal<SessionServiceEvent<ExampleUser>, NoError>.pipe()
+    }
+    
+    public func logIn(email: Email, _ password: String) -> SignalProducer<ExampleUser, SessionServiceError> {
+        sleep(2)
```
guidomb: You should use `dispatch_after` instead of sleep.

Authentication/RecoverPassword/RecoverPasswordController.swift
guidomb: Same with anything related with the recover password functionality.

Authentication/AuthenticationBootstrapper.swift
```diff
@@ -12,8 +12,9 @@ import Foundation
     Bootstrapper to start the application.
     Takes care of starting the authentication process before the main View Controller of the app when necessary,
     and after the user logs out.
+    The authentication process includes login, signup and recover password logic.
 */
-public class AuthenticationBootstraper<User: UserType, SessionService: SessionServiceType where SessionService.User == User> {
+public class AuthenticationBootstrapper<User: UserType, SessionService: SessionServiceType where SessionService.User == User> {
```
guidomb: We need to add an extension point (a method) to be able to set the view delegate without having to override the `createLoginView` method. Maybe the `ApplicationBootstrapper` could have a `var loginViewDelegate: LoginViewDelegate? = .None` property. What do you think?

guidomb:
@danielaRiesgo Remember to add a commit for each comment.

Authentication/LogIn/View/LoginController.swift
danielaRiesgo: The thing is if the user instead of presenting the register controller wants to directly change the rootviewcontroller of the window. The delegate doesn't have the window. Maybe we could have a mutable property that changes when register is selected and set the authentication bootstrapper to listen to it, and use the delegate for injecting more behaviour. Nevertheless, i believe there should be a default everything at all steps, even if it is an "empty" delegate.

Authentication/AuthenticationBootstrapper.swift
danielaRiesgo: We can have a createLoginViewDelegate to continue with the pattern, and setting it to the view before return it.
I don't think there's any "nice" way, we will end up having so many create methods or public vars. So any of the two are fine with me.

Authentication/LogIn/View/LoginController.swift
guidomb: That is not true because somebody needs to provide the window to the bootstrapper so that same object could have access to the window and pass it to the delegate.

Authentication/AuthenticationBootstrapper.swift
guidomb: In this case I think it makes more sense to have a property. Because this will be the most common way to configure the view and by exposing a property we avoid having to inherit from the bootstrapper to customize how the view will look like.

fedelcar:
This will fix the issue with font in the password text field when showing / hidding the password. (Code in objc from another project)

```objc 
 self.passwordTextField.enabled = NO;
self.passwordTextField.secureTextEntry = !self.passwordTextField.secureTextEntry;
 self.passwordTextField.enabled = YES;
 [self.passwordTextField setFont:[UIFont systemFontOfSize:14]];
```

Authentication/AuthenticationBootstrapper.swift
danielaRiesgo: So then we can do it with the controller delegate too. We leave the delegates as properties to be able to customize some changes without inheriting, but if they are so many changes or important ones, the user will need to inherit and override. Sounds good?

Authentication/LogIn/View/LoginController.swift
danielaRiesgo: So I should create the disposables in the init?

AuthenticationDemo/ExampleSessionService.swift
danielaRiesgo: But with dispatch_after you can't return a value

Authentication/LogIn/View/LoginView.swift
```diff
@@ -167,11 +167,22 @@ public final class LoginView: UIView, LoginViewType {
         }
     }
     
+    public var showPassword: Bool = false {
+        didSet {
+            passwordTextField.enabled = false
```
guidomb: Could you please add a comment explaining why this is needed?

Authentication/LogIn/View/LoginController.swift
```diff
@@ -153,7 +153,8 @@ private extension LoginController {
         loginView.logInButton.setTitle(_viewModel.loginButtonTitle, forState: .Normal)
         loginView.logInButton.rex_pressed.value = _viewModel.logInCocoaAction
         loginView.logInButtonEnabled = false
-        loginView.logInButton.rex_enabled.signal.observeNext { [unowned self] in self.loginView.logInButtonEnabled = $0 }
+        loginView.logInButton.rex_enabled.signal.observeNext { [unowned self] in
```
guidomb: Undo this change

guidomb:
Let's revert [this](https://github.com/Wolox/wolmo-authentication-ios/pull/2/commits/f59478779febe75fc79dda95f1aef833e998c2e5) commit

guidomb:
I've been discussing with the rest of the team and we want to minimize configuration setup looking this from the consumer side. What I am trying to say is that we by default user's won't be able to change the transition and the classes for each controller (unless they inherit from the AuthenticationConfiguration).

What I don't want is to make this super complex just to let the user configure how the transition will be made from the login controller to the register controller.

I would like to have something like the following setup:

```swift
protocol AuthenticationConfiguration {

  var loginControllerDelegate: LoginControllerDelegate { get }
  var loginViewDelegate: LoginViewDelegate { get }
  
  var registerControllerDelegate: RegisterControllerDelegate { get }
  var registerViewDelegate: RegisterViewDelegate { get }
  
}

extension AuthenticationConfiguration {

  var loginControllerDelegate: LoginControllerDelegate { return DefaultLoginControllerDelegate() }
  var loginViewDelegate: LoginViewDelegate { return DefaultLoginViewDelegate() }

  var registerControllerDelegate: RegisterControllerDelegate { return DefaultRegisterControllerDelegate() }
  var registerViewDelegate: RegisterViewDelegate { return DefaultRegisterViewDelegate() }

}

final class DefaultAuthenticationConfiguration: AuthenticationConfiguration {}

final class CustomLoginViewDelegate: LoginViewDelegate {

  func configureView(view: LoginViewType) {
    view.emailTextField.color = UIColor.redColor()
  }

}

final class CustomAuthenticationConfiguration: AuthenticationConfiguration {

  var loginViewDelegate: LoginViewDelegate { return CustomLoginViewDelegate() }

}

AuthenticationBootstrapper(window: window, configuration: CustomAuthenticationConfiguration()) {
  window.rootViewController = UINavigationController(rootViewController: MainViewController())
}
```

danielaRiesgo:
Travis failing because of configuration errors:

```
$ rvm use $(< .ruby-version) --install --binary --fuzzy
ruby-2.3.0 is not installed - installing.
Searching for binary rubies, this might take some time.
Requested binary installation but no rubies are available to download, consider skipping --binary flag.
Gemset '' does not exist, 'rvm ruby-2.3.0 do rvm gemset create ' first, or append '--create'.
The command "rvm use $(< .ruby-version) --install --binary --fuzzy" failed and exited with 2 during .
Your build has been stopped.
```

danielaRiesgo:
About the configuration setup.
I like more the first option, because it not only minimizes the arguments in the loginController but it also gives more structure and "entry points" with clear responsabilities for the user.
But at the same time, I was thinking that you are right, transitions are the least of the problems. I don't think anyone would use any other option than presenting or changing the root view controller (at least for 95% of times I think). So if they are changing the transitions, they are most probably changing other things and already inheriting from AuthenticationBootstrapper.
So maybe were overworking or overthinking, and the last option is quicker and still enables configuration. Although I would prefer having the LoginControllerActionDelegate (or TransitionDelegate) to separate responsabilities (without the AuthenticationConfiguration).

guidomb:
@danielaRiesgo OK lets do the third option extracting the `TransitionDelegate` from the Configuration object. For the ruby issue, lets downgrade ruby to `2.2.2`. Do `rbenv local 2.2.2 && gem install bundler && rbenv rehash && bundle install`.

Authentication/AuthenticationBootstrapper.swift
```diff
@@ -145,16 +145,35 @@ public class AuthenticationBootstrapper<User: UserType, SessionService: SessionS
...
-            loginViewFactory: createLoginView,
-            delegate: createLoginControllerDelegate())
+            viewFactory: createLoginView,
+            transitionDelegate: self)
+        return LoginController(configuration: configuration)
     }
+    
+}
 
+public extension AuthenticationBootstrapper {
```
guidomb: When you conform to protocols in extension declare the protocol in the extension

```swift
public extension AuthenticationBootstrapper: LoginControllerTransitionDelegate
```

Authentication/AuthenticationBootstrapper.swift
```diff
@@ -145,16 +145,35 @@ public class AuthenticationBootstrapper<User: UserType, SessionService: SessionS
...
+    
+    public func loginControllerDidTapOnRegister(controller: LoginController) {
+        let registerController = createRegisterController()
+        if let navigationController = controller.navigationController {
+            navigationController.pushViewController(registerController, animated: true)
+        } else {
+            self._window.rootViewController = UINavigationController(rootViewController: registerController)
```
guidomb: Remove `self`

Authentication/AuthenticationBootstrapper.swift
```diff
@@ -145,16 +145,35 @@ public class AuthenticationBootstrapper<User: UserType, SessionService: SessionS
...
+        }
+    }
+    
+    public func loginControllerDidTapOnRecoverPassword(controller: LoginController) {
+        let recoverPasswordController = createRecoverPasswordController()
+        if let navigationController = controller.navigationController {
+            navigationController.pushViewController(recoverPasswordController, animated: true)
+        } else {
+            self._window.rootViewController = UINavigationController(rootViewController: recoverPasswordController)
```
guidomb: Remove `self`

Authentication/LogIn/View/LoginController.swift
```diff
@@ -139,28 +133,22 @@ private extension LoginController {
...
         if let passwordVisibilityButton = loginView.passwordVisibilityButton {
-            passwordVisibilityButton.rex_title <~ _viewModel.showPassword.producer.map { [unowned self] _ in
-                self._viewModel.passwordVisibilityButtonTitle
-            }
+            passwordVisibilityButton.rex_title <~ _viewModel.showPassword.producer.map { [unowned self] _ in self._viewModel.passwordVisibilityButtonTitle }
             passwordVisibilityButton.rex_pressed.value = _viewModel.togglePasswordVisibility.unsafeCocoaAction
-            _viewModel.showPassword.signal.observeNext { [unowned self] showPassword in
-                self.loginView.showPassword = showPassword
-            }
+            _viewModel.showPassword.signal.observeNext { [unowned self] showPassword in self.loginView.showPassword = showPassword }
```
guidomb: This could be changed to

```swift
_viewModel.showPassword.signal.observeNext { [unowned self]  in self.loginView.showPassword = $0 }
```

Authentication/LogIn/View/LoginView.swift
```diff
@@ -33,7 +33,7 @@ public protocol LoginViewType: Renderable {
     var registerButton: UIButton { get }
     var recoverPasswordButton: UIButton { get }
     
-    var activityIndicator: UIActivityIndicatorView { get }
+    var activityIndicator: UIActivityIndicatorView? { get }
```
guidomb: Lets delete this.

guidomb:
@danielaRiesgo You need to configure CarthageCache in travis.

Authentication/LogIn/View/LoginController.swift
```diff
@@ -227,20 +227,11 @@ extension LoginController {
...
     */
     func calculateTextFieldOffsetToMoveFrame(keyboardOffset: CGFloat) -> CGFloat {
-        let emailOffset = loginView.emailTextField.convertPoint(loginView.emailTextField.frame.origin, toView: self.view).y - 10
-        let passwordBottom = loginView.emailTextField.convertPoint(loginView.passwordTextField.frame.origin, toView: self.view).y + loginView.passwordTextField.frame.size.height
-        if (keyboardOffset + (passwordBottom - emailOffset)) <= self.view.frame.size.height {
-            return emailOffset
-        } else {
-            return _activeField.value!.convertPoint(_activeField.value!.frame.origin, toView: self.view).y - 10
-        }
+        return loginView.emailTextField.convertPoint(loginView.emailTextField.frame.origin, toView: self.view).y - 10
```
guidomb: remove this extra line

guidomb:
@danielaRiesgo Check that the login controller initializer has a problem with spacing

script/bootstrap
```diff
@@ -76,7 +76,7 @@ bootstrap_carthage ()
   then
     carthage_cmd="$carthage_cmd --no-use-binaries"
   fi
-
+  echo "Disabled cache: $DISABLE_CARTHAGE_CACHE"
```
guidomb: Lets revert this commit. I'm gonna release a version of CarthageCache with better logging output.

danielaRiesgo: I erased it in another commit.

Authentication/Login/Model/LoginColourPalette.swift
```diff
@@ -0,0 +1,36 @@
...
+//  LoginColourPalette.swift
+//  Authentication
+//
+//  Created by Daniela Riesgo on 3/31/16.
+//  Copyright © 2016 Wolox. All rights reserved.
+//
+
+import Foundation
+
+public struct LoginColourPalette {
```
guidomb: Lets use American english for consistency with the SDK. Rename Colour to Color

guidomb:
@danielaRiesgo `LoginViewConfiguration` should be a protocol (`LoginViewConfigurationType`) and we should define a protocol extension with the default configuration.

guidomb:
@danielaRiesgo Are you sure that the user of this lib will be able to configure the color pallet? I am not sure but I think that if you define an initializer with no parameters then you won't be able to create a configuration with a different pallet. 

I think that for convenience all the properties in the pallet struct should be var.

woloxci:

guidomb:
:+1:

PR #3, opened 2016-04-01
Register view backend

danielaRiesgo:
## Sumary ##
**Trello card**: https://trello.com/c/Pei9poRd/7-vista-registro

Following the same structure as in the login view, this PR adds the register/singup view backend structure (controller, view model and view), relating it to the bootstrapper.

## Known issues ##
Missing tests.
Terms and Services button has no action yet.


woloxci:

guidomb:
👍

guidomb:
I'm adding @FranDepascuali also as a reviewer

Authentication/AuthenticationBootstrapper.swift
```diff
@@ -89,7 +91,7 @@ public class AuthenticationBootstrapper<User: UserType, SessionService: SessionS
          Creates the LogInViewModel to use in the authentication process logic,
          with the LogInCredentialsValidator returned in the function createLogInCredentialsValidator.
 
-         - Returns: A login view model that controls the log in logic and comunicates with the session service.
+         - Returns: A login view model that controls the login logic and comunicates with the session service.
```
FranDepascuali: communicates*

Authentication/AuthenticationBootstrapper.swift
```diff
@@ -108,7 +110,7 @@ public class AuthenticationBootstrapper<User: UserType, SessionService: SessionS
     */
     public func createLoginView() -> LoginViewType {
         let view: LoginView = LoginView.loadFromNib()
```
FranDepascuali:
```swift
let view: LoginView = .loadFromNib()
```

Authentication/Configuration/AuthenticationViewConfiguration.swift
```diff
@@ -0,0 +1,31 @@
...
+//
+
+import Foundation
+
+public struct AuthenticationViewConfiguration {
+    
+    public let loginConfiguration: LoginViewConfigurationType
+    public let signupConfiguration: SignupViewConfigurationType
+    
+    public init() {
```
FranDepascuali: I think we don't need this init (we can combine it with the other one)

Authentication/Configuration/AuthenticationViewConfiguration.swift
```diff
@@ -0,0 +1,31 @@
...
+        loginConfiguration = DefaultLoginViewConfiguration()
+        signupConfiguration = DefaultSignupViewConfiguration()
+    }
+    
+    public init(loginViewConfiguration: LoginViewConfigurationType, signupViewConfiguration: SignupViewConfigurationType = DefaultSignupViewConfiguration()) {
```
FranDepascuali: `loginViewConfiguration: LoginViewConfigurationType = DefaultLoginViewConfiguration()` also

Authentication/Configuration/AuthenticationViewConfiguration.swift
```diff
@@ -0,0 +1,31 @@
...
+        loginConfiguration = loginViewConfiguration
+        signupConfiguration = signupViewConfiguration
+    }
+    
+    public init(signupViewConfiguration: SignupViewConfigurationType, loginViewConfiguration: LoginViewConfigurationType = DefaultLoginViewConfiguration()) {
```
FranDepascuali: We don't need this init

Authentication/Configuration/ColorPalette.swift
```diff
@@ -21,14 +21,18 @@ public protocol ColorPaletteType {
...
 }
 
 public extension ColorPaletteType {
     
-    public var logInButtonDisabled: UIColor { return UIColor(hexString: "#d8d8d8ff")! }
-    public var logInButtonEnabled: UIColor { return UIColor(hexString: "#f5a623ff")! }
-    public var logInButtonExecuted: UIColor { return UIColor(hexString: "#e78f00ff")! }
+    public var mainButtonDisabled: UIColor { return UIColor(hexString: "#d8d8d8ff")! }
+    public var mainButtonEnabled: UIColor { return UIColor(hexString: "#f5a623ff")! }
+    public var mainButtonExecuted: UIColor { return UIColor(hexString: "#e78f00ff")! }
```
FranDepascuali: shouldn't we extract this to a protocol for buttons? (which declares a color for each button state)?

Authentication/Configuration/ColorPalette.swift
```diff
@@ -21,14 +21,18 @@ public protocol ColorPaletteType {
...
     
     public var textfieldsError: UIColor { return UIColor(hexString: "#d0021bff")! }
     public var textfieldsNormal: UIColor { return UIColor.clearColor() }
```
FranDepascuali: `.clearColor()`  (without the `UIColor.`)

Authentication/Configuration/ColorPalette.swift
```diff
@@ -37,7 +41,10 @@ public extension ColorPaletteType {
     public var background: UIColor { return UIColor(hexString: "#efefefff")! }
     
     public var links: UIColor { return UIColor(hexString: "#0076ffff")! }
-
+    public var labels: UIColor { return UIColor.blackColor() }
```
FranDepascuali: `.blackColor()`

Authentication/Configuration/ColorPalette.swift
```diff
@@ -37,7 +41,10 @@ public extension ColorPaletteType {
...
+    public var textfieldText: UIColor { return UIColor.blackColor() }
```
FranDepascuali: `.blackColor()`

Authentication/Configuration/ColorPalette.swift
```diff
@@ -37,7 +41,10 @@ public extension ColorPaletteType {
...
+    public var passwordVisibilityButtonText: UIColor { return UIColor(hexString: "#0076ffff")! }
+    public var mainButtonText: UIColor { return UIColor.whiteColor() }
```
FranDepascuali: `.whiteColor()`

Authentication/Configuration/FontPalette.swift
```diff
@@ -28,7 +28,7 @@ public extension FontPaletteType {
     public var passwordVisibilityButton: UIFont { return UIFont.systemFontOfSize(14) }
```
FranDepascuali: delete `UIFont.`

Authentication/Configuration/FontPalette.swift
```diff
@@ -28,7 +28,7 @@ public extension FontPaletteType {
...
     public var links: UIFont { return UIFont.systemFontOfSize(14) }
```
FranDepascuali: delete `UIFont.`

Authentication/Configuration/FontPalette.swift
```diff
@@ -28,7 +28,7 @@ public extension FontPaletteType {
...
     public var labels: UIFont { return UIFont.systemFontOfSize(14) }
```
FranDepascuali: delete `UIFont.`

Authentication/Configuration/FontPalette.swift
```diff
@@ -28,7 +28,7 @@ public extension FontPaletteType {
...
-    public var loginButton: UIFont { return UIFont.boldSystemFontOfSize(16) }
+    public var mainButton: UIFont { return UIFont.boldSystemFontOfSize(16) }
```
FranDepascuali: delete `UIFont`

Authentication/Login/View/LoginConfiguration.swift
```diff
@@ -13,7 +13,7 @@ import Foundation
 */
 public protocol LoginControllerTransitionDelegate {
```
FranDepascuali: Shouldn't we notify `userDidLogin` here?

```swift
func didLogin(controller: LoginController, user: User)
```
or we are handling this logic in other place?

Authentication/Login/View/LoginConfiguration.swift
```diff
@@ -13,7 +13,7 @@ import Foundation
...
     
-    func loginControllerDidTapOnRegister(controller: LoginController)
+    func loginControllerDidTapOnSignup(controller: LoginController)
```
FranDepascuali: Couldn't we change this naming? I'd prefer `DidTapOnSignup(controller: LoginController)`. We are copying `UIKit` way of naming the delegates while they said in last WWDC that this naming would be improved.

Authentication/Login/View/LoginConfiguration.swift
```diff
@@ -13,7 +13,7 @@ import Foundation
...
     
     func loginControllerDidTapOnRecoverPassword(controller: LoginController)
```
FranDepascuali: Consequently, I would name this 
```swift
didTapOnRecoverPassword(controller: LoginController)
```swift

File: Authentication/Login/View/LoginController.swift [hunk 18]
```diff
@@ -157,9 +149,9 @@ private extension LoginController {
         loginView.logInButton.rex_pressed.value = _viewModel.logInCocoaAction
         loginView.logInButton.rex_enabled.signal.observeNext { [unowned self] in self.loginView.logInButtonEnabled = $0 }
         
-        loginView.registerLabel.text = _viewModel.registerLabelText
-        loginView.registerButton.setTitle(_viewModel.registerButtonTitle, forState: .Normal)
-        loginView.registerButton.setAction { [unowned self] _ in self._transitionDelegate.loginControllerDidTapOnRegister(self) }
+        loginView.signupLabel.text = _viewModel.signupLabelText
+        loginView.signupButton.setTitle(_viewModel.signupButtonTitle, forState: .Normal)
```
FranDepascuali: I think we could move the fixed strings (like buttons and labels texts) to the view protocol and have this default.

Authentication/Login/View/LoginView.swift
```diff
@@ -206,9 +206,9 @@ public final class LoginView: UIView, LoginViewType, NibLoadable {
         didSet {
             let color: UIColor
             if logInButtonEnabled {
-                color = delegate.colourPalette.logInButtonEnabled
+                color = delegate.colourPalette.mainButtonEnabled
             } else {
-                color =  delegate.colourPalette.logInButtonDisabled
+                color =  delegate.colourPalette.mainButtonDisabled
```
FranDepascuali:
```swift
let colourPalette = delegate.colourPalette
let color = logInButtonEnabled ? colourPalette.mainButtonEnabled : colourPalette.mainButtonDisabled
```

Authentication/Login/View/LoginView.swift
```diff
@@ -218,9 +218,9 @@ public final class LoginView: UIView, LoginViewType, NibLoadable {
         didSet {
             let color: UIColor
             if logInButtonPressed {
-                color =  delegate.colourPalette.logInButtonExecuted
+                color =  delegate.colourPalette.mainButtonExecuted
             } else {
-                color =  delegate.colourPalette.logInButtonEnabled
+                color =  delegate.colourPalette.mainButtonEnabled
```
FranDepascuali: let colourPalette = delegate.colourPalette
let color = logInButtonEnabled ? colourPalette.mainButtonExecuted : colourPalette.mainButtonEnabled

Authentication/Login/ViewModel/LoginViewModel.swift
```diff
@@ -129,12 +130,12 @@ public extension LoginViewModel {
...
-    var registerLabelText: String {
-        return "login-view.to-register-label".localized
+    var signupLabelText: String {
+        return "login-view.to-signup-label".localized
     }
     
-    var registerButtonTitle: String {
-        return "login-view-model.register-button-title".localized
+    var signupButtonTitle: String {
+        return "login-view-model.signup-button-title".localized
```
FranDepascuali: I think this should go into the view

Authentication/Login/ViewModel/LoginViewModel.swift
```diff
@@ -129,12 +130,12 @@ public extension LoginViewModel {
         return "login-view-model.login-button-title".localized
     }
     
-    var registerLabelText: String {
-        return "login-view.to-register-label".localized
+    var signupLabelText: String {
+        return "login-view.to-signup-label".localized
```
FranDepascuali: I think this should go into the view

Authentication/SessionServiceType.swift
```diff
@@ -0,0 +1,83 @@
...
+import ReactiveCocoa
+import enum Result.NoError
+
+
+public enum SessionServiceEvent<User: UserType> {
+    case LogIn(User)
+    case LogOut(User)
+    case SignUp(User)
+    case LogInError(SessionServiceError)
+    case RegistrationError(SessionServiceError)
```
FranDepascuali: This should be `SignUpError`

Authentication/SessionServiceType.swift
```diff
@@ -0,0 +1,83 @@
...
+        must take care of:
+            sending the user to the observers
+            sending the login event through the events signal and
+            updating the currentUser property.
+        If the credentials aren't valid, the SignalProducer returned
+        must take care of:
+            sending the error to the observer and
+            sending the login error event through the events signal.
+    */
+    func logIn(email: Email, _ password: String) -> SignalProducer<User, SessionServiceError>
```
FranDepascuali: @guidomb Why are we "unnaming" the password label? it should have the "password" label

guidomb: I guess at some point we decided not to use labels because that was the swift 2.2 convention but I Swift 2.3 this changes so we may need to use labels again.

Authentication/SessionServiceType.swift
```diff
@@ -0,0 +1,83 @@
...
+         must take care of:
+             sending the user to the observers
+             sending the sign up and the login event through the events signal and
+             updating the currentUser property.
+         If the credentials aren't valid, the SignalProducer returned
+         must take care of:
+             sending the error to the observer and
+             sending the singup error event through the events signal.
+    */
+    func signUp(name: String, _ email: Email, _ password: String) -> SignalProducer<User, SessionServiceError>
```
FranDepascuali: @guidomb Why are we "unnaming" the email and password label? it should have those labels. In this case particularly, we do want to require the second item to be an email and the third one to be a password. I know that in the implementation we can give it a label, but this allows us also to avoid giving a label in the implementation. 
So we could end with something like this: 
```swift
signUp(name: "", "", "")
```
And the user should know which parameter correspond with each one. 
The only place I saw this unnaming of parameters was in this post (https://www.natashatherobot.com/swift-making-the-associated-type-parameter-readable-in-protocols/) but it had a different usage (in there, we don't know what `thing` actually is, so it is correct to leave it under interpretation).

Authentication/SignUp/View/SignupController.swift
```diff
@@ -0,0 +1,230 @@
...
+
+public final class SignupController: UIViewController {
+    
+    private let _viewModel: SignupViewModelType
+    private let _signupViewFactory: () -> SignupViewType
+    private let _delegate: SignupControllerDelegate
+    
+    public lazy var signupView: SignupViewType = self._signupViewFactory()
+    
+    private lazy var _notificationCenter: NSNotificationCenter = NSNotificationCenter.defaultCenter()
```
FranDepascuali: why is this var?
Shouldn't it be 
```swift
private let _notificationCenter: NSNotificationCenter = .defaultCenter()
```

Authentication/SessionServiceType.swift
guidomb: The problem was we how the first label works in Swift 2.2 which has been fixed in Swift 3.0 We can go back to full labels.

Authentication/SignUp/View/SignupController.swift
```diff
@@ -0,0 +1,230 @@
...
+    private var _disposable = CompositeDisposable()
+    private let _keyboardDisplayed = MutableProperty(false)
+    private let _activeTextField = MutableProperty<UITextField?>(.None)
+
+    
+    init(viewModel: SignupViewModelType, signupViewFactory: () -> SignupViewType, delegate: SignupControllerDelegate = DefaultSignupControllerDelegate()) {
```
FranDepascuali: I think we should specify why this is internal or/and to label it `internal` to avoid confusion.

Authentication/SignUp/View/SignupController.swift
```diff
@@ -0,0 +1,230 @@
...
+    
+    public override func loadView() {
+        self.view = signupView.view
+    }
+    
+    
+    public override func viewDidLoad() {
+        signupView.render()
+        bindViewModel()
+        let tapRecognizer: UITapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(dismissKeyboard))
```
FranDepascuali: you don't need ```: UITapGestureRecognizer ```

Authentication/SignUp/View/SignupController.swift
```diff
@@ -0,0 +1,230 @@
...
+        self.view.addGestureRecognizer(tapRecognizer)
```
FranDepascuali: delete self

Authentication/SignUp/View/SignupController.swift
```diff
@@ -0,0 +1,230 @@
...
+    deinit {
+        _disposable.dispose()
+    }
+    
+    required public init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    public override func loadView() {
+        self.view = signupView.view
```
FranDepascuali: delete self

Authentication/SignUp/View/SignupController.swift
```diff
@@ -0,0 +1,230 @@
...
+        bindViewModel()
+        let tapRecognizer: UITapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(dismissKeyboard))
+        self.view.addGestureRecognizer(tapRecognizer)
+    }
+    
+}
+
+private extension SignupController {
+    
+    func bindViewModel() {
```
FranDepascuali: we should decide wether we name private or not functions inside a private extension. I think @danielaRiesgo and I agree to use it, wether @guidomb you don't like that. 
While doing code review in greenco, I am more convinced that the function should be private rather than without label. I think we should leave the least possible as ambigouos. However, it's your call @guidomb

Authentication/SignUp/View/SignupController.swift
```diff
@@ -0,0 +1,230 @@
...
+    }
+    
+    func bindButtons() {
+        signupView.signupButton.setTitle(_viewModel.signupButtonTitle, forState: .Normal)
+        signupView.signupButton.rex_pressed.value = _viewModel.signUpCocoaAction
+        signupView.signupButton.rex_enabled.signal.observeNext { [unowned self] in self.signupView.signupButtonEnabled = $0 }
+        
+        signupView.termsAndServicesLabel?.text = _viewModel.termsAndServicesLabelText
+        signupView.termsAndServicesButton.setTitle(_viewModel.termsAndServicesButtonTitle, forState: .Normal)
+        //signupView.termsAndServicesButton -> Presents the terms and services
```
FranDepascuali: Why do we have this commented?

Authentication/SignUp/View/SignupController.swift
```diff
@@ -0,0 +1,230 @@
...
+    }
+    
+}
+
+extension SignupController: UITextFieldDelegate {
+    
+    public func textFieldShouldReturn(textField: UITextField) -> Bool {
+        if textField == signupView.usernameTextField {
```
FranDepascuali: I find it cleaner to have an objc associated property `nextTextField`. When we have core, we could add it.
This would end up being
```swift
    textField.nextTextField?. becomeFirstResponder()
```

Authentication/SignUp/View/SignupController.swift
```diff
@@ -0,0 +1,230 @@
...
+        } else if textField == signupView.passwordTextField {
+            signupView.passwordTextFieldSelected = true
+        } else {
+            signupView.passwordConfirmationTextFieldSelected = true
+        }
+        _activeTextField.value = textField
+    }
+    
+    public func textFieldDidEndEditing(textField: UITextField) {
+        if textField == signupView.usernameTextField {
```
FranDepascuali: maybe we could add a selected associated property also

Authentication/SignUp/View/SignupController.swift
guidomb: Yeah lets use no problem for me.

Authentication/SignUp/View/SignupControllerDelegate.swift
```diff
@@ -0,0 +1,109 @@
...
+
+    func signupControllerDidPassPasswordConfirmationValidation(controller: SignupController)
+    
+    func signupController(controller: SignupController, didFailPasswordConfirmationValidationWithErrors errors: [String])
+    
+}
+
+extension SignupControllerDelegate {
+    
+    public func shouldDisplaySignupErrorWithAlert() -> Bool {
```
FranDepascuali: shouldn't this be a computed property?

Authentication/SignUp/ViewModel/SignupViewModel.swift
```diff
@@ -0,0 +1,159 @@
...
+    public var termsAndServicesLabelText: String {
+        return "signup-view-model.terms-and-services.label-text".localized
+    }
+    
+    public var termsAndServicesButtonTitle: String {
+        return "signup-view-model.terms-and-services.button-title".localized
+    }
+    public var signupButtonTitle: String {
+        return "signup-view-model.signup-button-title".localized
+    }
```
FranDepascuali: I think that all these texts are view things, and not view model things.

Authentication/Validators/TextInputValidatorType.swift
```diff
@@ -15,6 +15,10 @@ public enum ValidationResult {
         return .Invalid(errors: [errorMessage])
     }
     
+    public static func valid() -> ValidationResult {
```
FranDepascuali: This should ve a static var

FranDepascuali:
it is pretty big, I did my best

Authentication/SignUp/View/SignupView.swift
```diff
@@ -0,0 +1,106 @@
...
+    var usernameTextFieldSelected: Bool { get set }
+    var emailTextFieldValid: Bool { get set }
+    var emailTextFieldSelected: Bool { get set }
+    var passwordTextFieldValid: Bool { get set }
+    var passwordTextFieldSelected: Bool { get set }
+    var passwordConfirmationTextFieldValid: Bool { get set }
+    var passwordConfirmationTextFieldSelected: Bool { get set }
+    var signupButtonEnabled: Bool { get set }
+    var signupButtonPressed: Bool { get set }
+
```
FranDepascuali: The SignUpViewType and LoginViewType share some common characteristics. I propose to do this, what do you think @guidomb @danielaRiesgo ?

```swift
public protocol Renderable {
    
    var view: UIView { get }
    
    func render()
    
}

public protocol AuthenticationFormType {
    
    var emailLabel: UILabel? { get }
    var emailTextField: UITextField { get }
    var emailValidationMessageLabel: UILabel? { get }
    
    var passwordLabel: UILabel? { get }
    var passwordTextField: UITextField { get }
    var passwordValidationMessageLabel: UILabel? { get }
    var passwordVisibilityButton: UIButton? { get }
    
    var usernameLabel: UILabel? { get }
    var usernameTextField: UITextField? { get }
    var usernameValidationMessageLabel: UILabel? { get }
    
    var emailTextFieldValid: Bool { get set }
    var emailTextFieldSelected: Bool { get set }
}

protocol SignUpFormType: AuthenticationFormType {
    
    var passwordConfirmLabel: UILabel? { get }
    var passwordConfirmTextField: UITextField { get }
    var passwordConfirmValidationMessageLabel: UILabel? { get }
    var passwordConfirmVisibilityButton: UIButton? { get }
    
    var signupButton: UIButton { get }
    
    var signupErrorLabel: UILabel? { get }
    
    var termsAndServicesButton: UIButton { get }
    var termsAndServicesLabel: UILabel? { get }
    
    var passwordTextFieldValid: Bool { get set }
    var passwordTextFieldSelected: Bool { get set }
    
    var usernameTextFieldValid: Bool { get set }
    var usernameTextFieldSelected: Bool { get set }
    
    var passwordConfirmationTextFieldValid: Bool { get set }
    var passwordConfirmationTextFieldSelected: Bool { get set }
    
    var signupButtonEnabled: Bool { get set }
    var signupButtonPressed: Bool { get set }
}

public protocol LoginFormType: AuthenticationFormType {
    
    var view: UIView { get }
    
    var logoImageView: UIImageView { get }
    
    var logInButton: UIButton { get }
    var logInErrorLabel: UILabel? { get }
    
    var signUpLabel: UILabel { get }
    var signUpButton: UIButton { get }
    var recoverPasswordButton: UIButton { get }
    
    var logInButtonEnabled: Bool { get set }
    var logInButtonPressed: Bool { get set }
    var showPassword: Bool { get set }
    
}

typealias SignUpViewType = protocol<SignUpFormType, Renderable>

typealias LoginViewType = protocol<LoginFormType, Renderable>
```

guidomb: Cool Why not make the `SignupFormType` and the `LoginFormType` extend `Renderable` ? Any particular reason why those should be separated? Just curious, it's fine either way for me

FranDepascuali: I see `SignUpFormType` and `LoginFormType` both as an extension of the `AuthenticationFormType`, and by doing that, we assure that we are consistent for the things we are asking for in the `AuthenticationFormType`. For example, if we need an email for login, we are sure that we are also asking for it in `AuthenticationFormType`. I'm just saying this because I just realized that we gain by having both extend a common protocol.


For the `Renderable`, first I thought that it didn't need to extend `Renderable`. Now that I see it, both ```SignUpFormType: AuthenticationFormType, Renderable``` and ```typealias SignUpViewType = protocol<SignUpFormType, Renderable>``` are the same. The only thing is that (I don't know why) I assumed that the second way of writing gives a better notion of composability than the first one, because one (which happened to me) can see it as inheritance. But they are the same and it could be written both ways, maybe the first one is better so we avoid that extra line.

FranDepascuali: Now that I've said that last thing, I think that `LoginViewType` should be like this:
```swift
public protocol LoginViewType: LoginFormType, Renderable {

    var logoImageView: UIImageView { get }

}
```
As the logo image is not part of a form but a component of the view

guidomb: Yeah my argument in favor of making `AuthenticationFormType` is that I cannot imagine a use case were you wouldn't be able to return the container view for the form. I mean, all form's components should be contained in a parent view (other wise you won't be able to use them or at least display them), by making `AuthenticationFormType` extend `Renderable` we are assuming that the one implementing the protocol has access to the container view. Which I think is a pretty safe assumption to make. Because the one that implements this protocol should be in charge of laying out the components and the safest way I can think of doing that is by creating a container view and use autolayouts for the form components which will then be added as childs in the container view.

Authentication/Configuration/ColorPalette.swift
danielaRiesgo: I don't see the need for that.

Authentication/Login/View/LoginConfiguration.swift
danielaRiesgo: When the user logs in, the bootstrapper automatically opens your app's main view. Do you think someone may want to do something before that?
@guidomb

Authentication/Login/View/LoginController.swift
danielaRiesgo: The only thing I don't like is to have localized string scattered through more than one class. And as the password visibility button's title need information from view model, I think I would leave them all in the view model.
Or maybe make a function in the view to get the localized string from a bool or directly change the title by itself, and make the controller notify when the password visibility property changes.
I'm okey with the second one

Authentication/SignUp/View/SignupController.swift
danielaRiesgo: Should we receive it in the initializer? For both controllers to use the same.
 @guidomb

Authentication/SignUp/View/SignupController.swift
danielaRiesgo: I will add a comment although I think this will be said in the framework specifications, because the view models also have internal initializers.
Unless we want to make them public. What do you think? I think it's better to just let them use the bootstrapper to configure our internal classes instead of letting them create them.
@guidomb

Authentication/SignUp/View/SignupController.swift
danielaRiesgo: Because we haven't discussed yet how this would be. Maybe it's time to. @guidomb 
The terms and services should be a link and the button should take the user to the browser? Or something else?

Authentication/SignUp/View/SignupController.swift
danielaRiesgo: Could we already add core?

Authentication/SignUp/View/SignupController.swift
danielaRiesgo: The only thing is we should make a class or something to associate, because it's just a bool for now.

Authentication/Validators/TextInputValidatorType.swift
danielaRiesgo: I would like it to stay a function just to be consistent with the invalid function. But maybe it's just a whim of mine.

Authentication/SignUp/View/SignupView.swift
danielaRiesgo: I like it. I think all of these need a container view, but `Renderable` I think it represents a view that must be "configured" or rendered even after initialized to look okay. And I don't think the logo or the  labels and textfields need that.

danielaRiesgo: What I would do is make the Form types include all textfields and labels and maybe the main button. But leave the logo and the buttons to other screens to the LoginViewType or SignupViewType.

danielaRiesgo:
```swift
public protocol Renderable {
    
    var view: UIView { get }
    
    func render()
    
}

public extension Renderable where Self: UIView {
    
    var view: UIView {
        return self
    }
    
}


public protocol AuthenticationFormType {
    
    var emailLabel: UILabel? { get }
    var emailTextField: UITextField { get }
    var emailValidationMessageLabel: UILabel? { get }
    
    var emailTextFieldValid: Bool { get set }
    var emailTextFieldSelected: Bool { get set }
    
    var passwordLabel: UILabel? { get }
    var passwordTextField: UITextField { get }
    var passwordValidationMessageLabel: UILabel? { get }
    var passwordVisibilityButton: UIButton? { get }
    
    var passwordTextFieldValid: Bool { get set }
    var passwordTextFieldSelected: Bool { get set }
    var showPassword: Bool { get set }
    
}

public protocol LoginFormType: AuthenticationFormType {

    var logInButton: UIButton { get }
    var logInErrorLabel: UILabel? { get }
    
    var logInButtonEnabled: Bool { get set }
    var logInButtonPressed: Bool { get set }

}

public protocol SignupFormType: AuthenticationFormType {

    var usernameLabel: UILabel? { get }
    var usernameTextField: UITextField? { get }
    var usernameValidationMessageLabel: UILabel? { get }
    
    var usernameTextFieldValid: Bool { get set }
    var usernameTextFieldSelected: Bool { get set }

    var passwordConfirmLabel: UILabel? { get }
    var passwordConfirmTextField: UITextField { get }
    var passwordConfirmValidationMessageLabel: UILabel? { get }
    var passwordConfirmVisibilityButton: UIButton? { get }
    
    var passwordConfirmationTextFieldValid: Bool { get set }
    var passwordConfirmationTextFieldSelected: Bool { get set }
    var showConfirmPassword: Bool { get set }
    
    var signupButton: UIButton { get }
    var signupErrorLabel: UILabel? { get }
    
    var signupButtonEnabled: Bool { get set }
    var signupButtonPressed: Bool { get set }
    
    var termsAndServicesButton: UIButton { get }
    var termsAndServicesLabel: UILabel? { get }
    
}

public protocol LoginViewType: Renderable, LoginFormType {
    
    var logoImageView: UIImageView { get }
    
    var signupLabel: UILabel { get }
    var signupButton: UIButton { get }
    var recoverPasswordButton: UIButton { get }
    
}

public protocol SignupViewType: Renderable, SignupFormType { }
```

woloxci:

Authentication/SignUp/View/SignupView.swift
guidomb: Yeah I like this approach too. If everyone agrees lets go with this one.

Authentication/Login/View/LoginConfiguration.swift
guidomb: If somebody else wants to know when a user has logged in they should use the service for that. This is intended only to customize what this particular screen should do after a user has logged in. We use it to continue with the flow.

Authentication/Validators/TextInputValidatorType.swift
guidomb: Are there any rules defined for such things in either the github styel guid or swift's?

Authentication/Login/View/LoginConfiguration.swift
danielaRiesgo: But now I was thinking. We are leaving all the transitions to the TransitionDelegate, even from `LogIn` to `SignUp`, so maybe we could leave the transition from the LogIn to the main screen of the app (for cases were maybe the user of the framework wants to send the user when logged in or signup to a welcome page and then to the main screen, or I don't know). And instead of making the bootstrapper hear the `SessionService`'s events to act, make it implement a method `userDidLogInSuccessfully` from the TransitionDelegate protocol.

Authentication/Login/View/LoginConfiguration.swift
guidomb: I would change this completely when we add support for swift 3. I think a better name would be `onRegister` and `onSignup`. That is how I would it name it in any other language if it wasn't for UIKit weird conventions.

Authentication/SignUp/View/SignupView.swift
danielaRiesgo: We could also do something similar with the view models.

Authentication/Login/View/LoginConfiguration.swift
guidomb: That could be an option too. My point is that this delegate should be only used to decide what to do with the login screen. Any other "independent" action that needs to be performed when the user logins should be done using the session service.

danielaRiesgo: So do I add that method `userDidLogIn`? The only thing I don't like with that, is that the `SessionService`'s events signal becomes somewhat useless since we will only use the `LogOut` event to redirect to login, which is great, but I think we shouldn't demand many events from the service if we will only use one.

guidomb:
I think we are ready to merge this 👍

Authentication/Validators/TextInputValidatorType.swift
danielaRiesgo: This issue is not resolved.

Authentication/SignUp/View/SignupController.swift
danielaRiesgo: This issue is not resolved.

Authentication/SignUp/View/SignupController.swift
danielaRiesgo: This issue is not resolved.

Authentication/SignUp/View/SignupController.swift
danielaRiesgo: This issue is not resolved.

Authentication/SignUp/View/SignupController.swift
danielaRiesgo: This issue is not resolved.

Authentication/Validators/TextInputValidatorType.swift
guidomb: If we want to do what the new Swift guidelines are doing this should probably be a readonly property.

Authentication/Login/View/LoginConfiguration.swift
danielaRiesgo: This issue is not resolved.

Authentication/SignUp/View/SignupController.swift
guidomb: In theory if you want to follow dependency injection all dependencies should be passed through the constructor / initializer as interfaces / protocol so they can be mocked when testing. That being said I've never found my self in a situation that I wanted to use a different instance of `NSNotificationCenter` rather than the default one. So, I would say leave it as an internal implementation detail and use a `let`.

danielaRiesgo:
I added a `This issue is not resolved.` comment in the things that still need discussion or closure.

Authentication/SignUp/View/SignupController.swift
guidomb: There are 2 things here. The first one is that I think that we should use `internal` explicitly everywhere. Maybe for our internal project we should discuss it with the rest of the team but the OSS projects I would follow the GitHub guidelines quite close. I think it makes everything way more clear to explicitly state that this method or class is for internal use only.

In this particular case I agree with @danielaRiesgo. The initializer should be `internal` because the controller is not intended for customization because is just glue code between the view and the view model. Unlike the view or view model where there are protocols and if you want to have a different implementation of the view for example is just a matter of creating a new class and conforming to the protocol.

Authentication/SignUp/View/SignupController.swift
guidomb: This link should display a view controller that has a web view that will display the terms and service which is just an HTML page with some CSS.

Authentication/SignUp/View/SignupController.swift
guidomb: @FranDepascuali what do you mean by making an associated property?

Authentication/SignUp/View/SignupController.swift
danielaRiesgo: Okay, great. Could that be in another PR?

guidomb: For sure

Authentication/SignUp/View/SignupController.swift
guidomb: Yeah lets add that into another PR so we can close this one.

Authentication/SignUp/View/SignupView.swift
```diff
@@ -8,49 +8,49 @@
...
-    public var passwordTextFieldSelected: Bool = false
-    public var showPassword: Bool = false
-    public var passwordConfirmationTextFieldValid: Bool = false
-    public var passwordConfirmationTextFieldSelected: Bool = false
-    public var showConfirmPassword: Bool = false
-    public var signupButtonEnabled: Bool = false
-    public var signupButtonPressed: Bool = false
-    
-    public func render() {
+    internal var usernameLabel: UILabel? { return .None }
```
guidomb: Why did you change all this to `internal`?

danielaRiesgo: Because this view is the internal one, the "default implementation" of the protocol. If the user wants to use this, it should be configured through the bootstrapper, not directly.

Authentication/Login/View/LoginConfiguration.swift
danielaRiesgo: In the protocol `SessionServiceType` which the user must always implement, we ask that they have a property `var events: Signal<SessionServiceEvent<User>, NoError> { get }`, where 
```swift
public enum SessionServiceEvent<User: UserType> {
    case LogIn(User)
    case LogOut(User)
    case SignUp(User)
    case LogInError(SessionServiceError)
    case SignUpError(SessionServiceError)
}
```
And the only place we use this signal is in the `AuthentiationBootstrapper`:
```swift
sessionService.events.observeNext { [unowned self] event in
            switch event {
            case .LogIn(_): self._window.rootViewController = self._mainViewControllerFactory()
            case .LogOut(_): self._window.rootViewController = UINavigationController(rootViewController: self.createLoginController())
            default: break
            }
        }
```

And if we now add `userDidLog` in the `LogInTransitionDelegate` protocol, we wouldn't even need the `LogIn`event.

The user when seeing that he must conform to that protocol, I think he/she will think he must send all that events, when we don't really need those event notifications from him/her.

woloxci:

Authentication/Login/View/LoginController.swift
```diff
@@ -203,9 +191,9 @@ extension LoginController: UITextFieldDelegate {
     
 }
 
-extension LoginController {
+internal extension LoginController {
```
FranDepascuali: This should be private.

Authentication/Login/View/LoginController.swift
```diff
@@ -251,26 +239,16 @@ extension LoginController {
...
         textfield at the top of the screen.
     */
-    func calculateTextFieldOffsetToMoveFrame(keyboardOffset: CGFloat) -> CGFloat {
+    private func calculateTextFieldOffsetToMoveFrame(keyboardOffset: CGFloat) -> CGFloat {
         return loginView.emailTextField.convertPoint(loginView.emailTextField.frame.origin, toView: self.view).y - 10
     }
     
-    func dismissKeyboard(sender: UITapGestureRecognizer) {
+    // Internal to avoid using @objc label for using it in selector.
+    internal func dismissKeyboard(sender: UITapGestureRecognizer) {
```
FranDepascuali: this works? Having a protocol conformance in an extension? Anyway, you could move this to another extension and leave this extension as private. 
(Though I prefer to label this private and have the @objc)

Authentication/SignUp/View/SignupView.swift
guidomb: OK makes sense.

Authentication/Login/View/LoginView.swift
```diff
@@ -129,132 +115,183 @@ public final class LoginView: UIView, LoginViewType, NibLoadable {
...
-            emailErrorsHeightConstraint.constant = 0
-            emailErrorsHeightConstraint.active = true
+    private func passwordTextFieldSelectedWasSet() {
+        if passwordTextFieldSelected {
+            passwordTextFieldAndButtonViewOutlet.layer.borderColor = delegate.colorPalette.textfieldsSelected.CGColor
             passwordErrorsHeightConstraint.constant = 0
             passwordErrorsHeightConstraint.active = true
+        } else {
+            let valid = passwordTextFieldValid
+            passwordTextFieldValid = valid
```
FranDepascuali: I don't understand this two lines

Authentication/Login/View/LoginController.swift
guidomb: If you make it private you won't be able to use it in the bootstrapper.

Authentication/Login/View/LoginViewConfiguration.swift
```diff
@@ -11,27 +11,26 @@ import Foundation
...
     public let fontPalette: FontPaletteType
     
     public init() {
         logoImage = .None
-        colourPalette = DefaultColorPalette()
+        colorPalette = DefaultColorPalette()
         fontPalette = DefaultFontPalette()
     }
     
     public init(logoImage: UIImage?) {
```
FranDepascuali: I would just have one init method:
```swift
public init(logoImage: UIImage? = .None, 
                 colorPalette: ColorPaletteType = DefaultColorPalette(), 
                 fontPalette: FontPaletteType = DefaultFontPalette()) {

    self.logoImage = logoImage
         self.colorPalette = colorPalette
         self.fontPalette = fontPalette	       
}
```

Authentication/Login/View/LoginViewDelegate.swift
```diff
@@ -32,19 +32,19 @@ public extension LoginViewDelegate {
 public final class DefaultLoginViewDelegate: LoginViewDelegate {
     
     private let _configuration: LoginViewConfigurationType
-    
-    public let colourPalette: ColorPaletteType
+
+    public let colorPalette: ColorPaletteType
     public let fontPalette: FontPaletteType
 
     init() {
```
FranDepascuali: I would have one init method:
```swift
    public init(configuration: LoginViewConfigurationType) {
        _configuration = configuration
    }
```

Authentication/Login/View/LoginViewDelegate.swift
```diff
@@ -32,19 +32,19 @@ public extension LoginViewDelegate {
 public final class DefaultLoginViewDelegate: LoginViewDelegate {
     
     private let _configuration: LoginViewConfigurationType
-    
-    public let colourPalette: ColorPaletteType
+
+    public let colorPalette: ColorPaletteType
     public let fontPalette: FontPaletteType
```
FranDepascuali: we can delete these properties and have them as computed properties from the configuration

Authentication/Login/ViewModel/LoginViewModel.swift
```diff
@@ -22,21 +22,11 @@ public protocol LoginViewModelType {
     var passwordValidationErrors: AnyProperty<[String]> { get }
     var showPassword: MutableProperty<Bool> { get }
     
-    var togglePasswordVisibility: Action<AnyObject, Bool, NoError> { get }
+    var togglePasswordVisibilityCocoaAction: CocoaAction { get }
```
FranDepascuali: Why did we change this to a CocoaAction?

Authentication/Login/ViewModel/LoginViewModel.swift
```diff
@@ -71,12 +61,13 @@ public final class LoginViewModel<User: UserType, SessionService: SessionService
...
     
     
-    public private(set) lazy var togglePasswordVisibility: Action<AnyObject, Bool, NoError> = {
+    private lazy var _togglePasswordVisibility: Action<AnyObject, Bool, NoError> = {
         return Action { [unowned self] _ in
             self.showPassword.value = !self.showPassword.value
             return SignalProducer(value: self.showPassword.value).observeOn(UIScheduler())
         }
     }()
+    public var togglePasswordVisibilityCocoaAction: CocoaAction { return _togglePasswordVisibility.unsafeCocoaAction }
```
FranDepascuali: add \n before togglePasswordVisibilityCocoaAction

Authentication/SignUp/View/SignupController.swift
FranDepascuali: I thought that we could have a boolean objc associated property "selected" to avoid having it in the view (just an idea though).
This would be reduced to 
```swift
public func textFieldDidEndEditing(textField: UITextField) {
    textField.selected = false
    _activeTextField.value = .None
}
```

Authentication/SignUp/View/SignupView.swift
```diff
@@ -0,0 +1,106 @@
...
+    internal var passwordConfirmValidationMessageLabel: UILabel? { return .None }
+    internal var passwordConfirmVisibilityButton: UIButton? { return .None }
+    
+    internal var signupButton: UIButton { return UIButton() }
+    internal var signupErrorLabel: UILabel? { return .None }
+    
+    internal var termsAndServicesButton: UIButton { return UIButton() }
+    internal var termsAndServicesLabel: UILabel? { return .None }
+    
+    internal var usernameTextFieldValid: Bool = false
```
FranDepascuali: delete `: Bool`

Authentication/SignUp/View/SignupView.swift
```diff
@@ -0,0 +1,106 @@
...
+    internal var usernameTextFieldSelected: Bool = false
```
FranDepascuali: delete `: Bool`

Authentication/SignUp/View/SignupView.swift
```diff
@@ -0,0 +1,106 @@
...
+    internal var emailTextFieldValid: Bool = false
```
FranDepascuali: delete `: Bool`

Authentication/SignUp/View/SignupView.swift
```diff
@@ -0,0 +1,106 @@
...
+    internal var emailTextFieldSelected: Bool = false
```
FranDepascuali: delete `: Bool`

Authentication/SignUp/View/SignupView.swift
```diff
@@ -0,0 +1,106 @@
...
+    internal var passwordTextFieldValid: Bool = false
```
FranDepascuali: delete `: Bool`

Authentication/SignUp/View/SignupView.swift
```diff
@@ -0,0 +1,106 @@
...
+    internal var passwordTextFieldSelected: Bool = false
```
FranDepascuali: delete `: Bool`

Authentication/SignUp/View/SignupView.swift
```diff
@@ -0,0 +1,106 @@
...
+    internal var showPassword: Bool = false
```
FranDepascuali: delete `: Bool`

Authentication/SignUp/View/SignupView.swift
```diff
@@ -0,0 +1,106 @@
...
+    internal var passwordConfirmationTextFieldValid: Bool = false
```
FranDepascuali: delete `: Bool`

Authentication/SignUp/View/SignupView.swift
```diff
@@ -0,0 +1,106 @@
...
+    internal var passwordConfirmationTextFieldSelected: Bool = false
```
FranDepascuali: delete `: Bool`

Authentication/SignUp/View/SignupView.swift
```diff
@@ -0,0 +1,106 @@
...
+    internal var showConfirmPassword: Bool = false
```
FranDepascuali: delete `: Bool`

Authentication/SignUp/View/SignupView.swift
```diff
@@ -0,0 +1,106 @@
...
+    internal var signupButtonEnabled: Bool = false
```
FranDepascuali: delete `: Bool`

Authentication/SignUp/View/SignupView.swift
```diff
@@ -0,0 +1,106 @@
...
+    internal var signupButtonPressed: Bool = false
```
FranDepascuali: delete `: Bool`

Authentication/SignUp/View/SignupViewConfiguration.swift
```diff
@@ -0,0 +1,28 @@
...
+}
+
+public final class DefaultSignupViewConfiguration: SignupViewConfigurationType {
+    
+    public let colourPalette: ColorPaletteType
+    public let fontPalette: FontPaletteType
+    
+    public init() {
+        colourPalette = DefaultColorPalette()
+        fontPalette = DefaultFontPalette()
```
FranDepascuali: I would change this init for receiving the palettes as dependencies (and having them with default values)

Authentication/SignUp/View/SignupViewDelegate.swift
```diff
@@ -0,0 +1,110 @@
...
+    }
+    
+}
+
+public final class DefaultSignupViewDelegate: SignupViewDelegate {
+    
+    public let colourPalette: ColorPaletteType
+    public let fontPalette: FontPaletteType
+    
+    public init() {
```
FranDepascuali: I would change this init for receiving the palettes as dependencies (and having them with default values)

Authentication/SignUp/View/SignupViewDelegate.swift
```diff
@@ -0,0 +1,110 @@
...
+        colourPalette = DefaultColorPalette()
+        fontPalette = DefaultFontPalette()
+    }
+    
+    public init(configuration: SignupViewConfigurationType) {
```
FranDepascuali: I would make this a convenience init and use the other init

AuthenticationDemo/ExampleSessionService.swift
```diff
@@ -33,29 +33,31 @@ public final class ExampleSessionService: SessionServiceType {
...
+                return logInFailureSignalProducer(dispatchTime)
             }
         } else {
-            return failureSignalProducer(dispatchTime)
+            return logInFailureSignalProducer(dispatchTime)
         }
     }
     
-    private func successSignalProducer(user: ExampleUser, dispatchTime: dispatch_time_t) -> SignalProducer<ExampleUser, SessionServiceError> {
+    private func logInSuccessSignalProducer(user: ExampleUser, dispatchTime: dispatch_time_t) -> SignalProducer<ExampleUser, SessionServiceError> {
```
FranDepascuali: I would not name this with `SignalProducer`, just `logInSuccesful(...) -> ...`

AuthenticationDemo/ExampleSessionService.swift
```diff
@@ -67,14 +69,45 @@ public final class ExampleSessionService: SessionServiceType {
         })
     }
     
-    private func failureSignalProducer(dispatchTime: dispatch_time_t) -> SignalProducer<ExampleUser, SessionServiceError> {
+    private func logInFailureSignalProducer(dispatchTime: dispatch_time_t) -> SignalProducer<ExampleUser, SessionServiceError> {
```
FranDepascuali: I would not name this with SignalProducer, just logInFailure(...) -> ....
You could leave it like this though if you prefer it this way

AuthenticationDemo/ExampleSessionService.swift
```diff
@@ -67,14 +69,45 @@ public final class ExampleSessionService: SessionServiceType {
...
+            } else {
+                let user = ExampleUser(email: email.raw, password: password)
+                return signUpSuccessSignalProducer(user, dispatchTime: dispatchTime)
+            }
+        } else {
+            return signUpFailureSignalProducer(dispatchTime)
+        }
+    }
+    
+    private func signUpSuccessSignalProducer(user: ExampleUser, dispatchTime: dispatch_time_t) -> SignalProducer<ExampleUser, SessionServiceError> {
```
FranDepascuali: I would not name this with `SignalProducer`, just `signUpSuccessful(...) -> ....`
You could leave it like this though if you prefer it this way

AuthenticationDemo/ExampleSessionService.swift
```diff
@@ -67,14 +69,45 @@ public final class ExampleSessionService: SessionServiceType {
...
+        return SignalProducer<ExampleUser, SessionServiceError> { observer, disposable in
+            dispatch_after(dispatchTime, dispatch_get_main_queue()) {
+                observer.sendNext(user)
+                observer.sendCompleted()
+            }
+        }
+    }
+    
+    private func signUpFailureSignalProducer(dispatchTime: dispatch_time_t) -> SignalProducer<ExampleUser, SessionServiceError> {
```
FranDepascuali: I would not name this with `SignalProducer`, just `signUpFailure(...) -> ....`
You could leave it like this though if you prefer it this way

AuthenticationTests/LogIn/LoginControllerTests.swift
```diff
@@ -58,7 +63,7 @@ class LoginViewControllerSpec: QuickSpec {
                     sessionService = OneMyUserSessionService(email: Email(raw: "myuser@mail.com")!, password: "password", name: "MyUser")
                     loginVM = LoginViewModel(sessionService: sessionService)
                     loginV = LoginView()
-                    loginC = LoginController(viewModel: loginVM, loginViewFactory: { return loginV }, onRegister: { _ in }, onLoginError: .None)
+                    loginC = LoginController(viewModel: loginVM, loginViewFactory: { return loginV }, onSignup: { _ in }, onLoginError: .None)
                     loginC.viewDidLoad()
                     loginV.passwordVisibilityButton!.enabled = true
                     loginV.passwordVisibilityButton!.userInteractionEnabled = true
```
FranDepascuali: Why is this commented?

AuthenticationTests/LogIn/LoginControllerTests.swift
```diff
@@ -33,6 +33,11 @@ class LoginViewControllerSpec: QuickSpec {
                 loginC = LoginController(configuration: configuration)
```
FranDepascuali: This should be `loginController`

Authentication/Login/View/LoginController.swift
danielaRiesgo: But the things in this extension aren't used in the bootstrapper.

Authentication/Login/View/LoginView.swift
danielaRiesgo: It's to trigger the effect that is already defined in the didSet of the `passwordTextFieldValid ` property. And if done in one line swiftlint complains, if I'm not mistaken.

Authentication/Login/ViewModel/LoginViewModel.swift
danielaRiesgo: To be consistent with the log in action which is a CocoaAction. And at that time, if I'm not mistaken, it was because we wanted to ask the minimum necessary from the user.

Authentication/SignUp/View/SignupViewConfiguration.swift
danielaRiesgo: I think the default configuration must have the default palette's because if not it's not a default one. If not we could change its name maybe for `SignupViewConfiguration` and make it a struct with contains the palettes received.

AuthenticationTests/LogIn/LoginControllerTests.swift
danielaRiesgo: Because the `sendActionsForControlEvents` method don't work if there isn't a host application. Nevertheless, there is a card for this.

FranDepascuali:
I think that we can already merge this PR and we can discuss those details later

FranDepascuali:
👍

guidomb:
👍 Awesome!!! Nice work guys!! 🎉 🎉 🎉

PR #10, opened 2016-07-15
Register view frontend

danielaRiesgo:
## Sumary ##
**Trello card**: https://trello.com/c/BuvnKgwk/3-vista-sign-up

Adding the sign up default view with all visual "effects".
Adding what necessary to controller for transitions and binding with view objects that were missing.

## Test Plan ##

  * Open the app
  * Try to login: 
       It will only succeed if using email "example@mail.com" and password "password". And take you to a blank view.
  * Press Signup button to go to signup.
  * Try to signup:
       It will only succeed if using email "example@mail.com" and password "password". And take you to a blank view.
  * Press Login button to go to login.

### Updated Login and Signup
Version 2: with stack views
signup view and login v2

## Known issues ##
Terms and Services pending for another PR.
Missing tests.


woloxci:

Authentication/Login/View/LoginView.swift
```diff
@@ -27,80 +27,56 @@ internal final class LoginView: UIView, LoginViewType, NibLoadable {
...
         }
     }
     
-    internal var logInErrorLabel: UILabel? { return logInErrorLabelOutlet }
-    @IBOutlet weak var logInErrorLabelOutlet: UILabel! {
-        didSet {
-            logInErrorLabelOutlet.text = " "
-        }
-    }
+    internal var logInErrorLabel: UILabel? { return .None }
```
FranDepascuali: I think you don't need the `return .None`, isn't it implicit?

Authentication/Login/View/LoginView.swift
```diff
@@ -27,80 +27,56 @@ internal final class LoginView: UIView, LoginViewType, NibLoadable {
     
     internal var emailTextField: UITextField { return emailTextFieldOutlet }
     @IBOutlet weak var emailTextFieldOutlet: UITextField! {
-        didSet {
-            emailTextFieldOutlet.placeholder = emailPlaceholderText
-        }
+        didSet { emailTextFieldOutlet.placeholder = emailPlaceholderText }
```
FranDepascuali: I don't like this change :(

Authentication/SignUp/View/SignupConfiguration.swift
```diff
@@ -0,0 +1,62 @@
...
+ Protocol for handling transition events occured during signup.
+ */
+public protocol SignupControllerTransitionDelegate {
+    
+    func onLogin(controller: SignupController)
+    
+}
+
+/**
+ Class for configuring the singup controller.
```
FranDepascuali: signup

Authentication/SignUp/View/SignupConfiguration.swift
```diff
@@ -0,0 +1,62 @@
...
+ Includes all information required:
+ view factory method,
+ view model,
+ signup controller delegate and
+ signup controller transition delegate.
+ */
+public final class SignupControllerConfiguration {
+    
+    var viewModel: SignupViewModelType
```
FranDepascuali: why is this var? 
I also think we should declare its accessibility, as as simple sight I don't know if its meant to be public or internal.

Authentication/SignUp/View/SignupConfiguration.swift
```diff
@@ -0,0 +1,62 @@
...
+    var viewFactory: () -> SignupViewType
```
FranDepascuali: why is this var? 
I also think we should declare its accessibility, as as simple sight I don't know if its meant to be public or internal.

Authentication/SignUp/View/SignupConfiguration.swift
```diff
@@ -0,0 +1,62 @@
...
+    var delegate: SignupControllerDelegate
```
FranDepascuali: why is this var? 
I also think we should declare its accessibility, as as simple sight I don't know if its meant to be public or internal.

Authentication/SignUp/View/SignupConfiguration.swift
```diff
@@ -0,0 +1,62 @@
...
+    var transitionDelegate: SignupControllerTransitionDelegate
```
FranDepascuali: why is this var? 
I also think we should declare its accessibility, as as simple sight I don't know if its meant to be public or internal.

Authentication/SignUp/View/SignupController.swift
```diff
@@ -6,14 +6,14 @@
 //  Copyright © 2016 Wolox. All rights reserved.
 //
 
-import Foundation
 import ReactiveCocoa
 
 public final class SignupController: UIViewController {
     
-    private let _viewModel: SignupViewModelType
+    private var _viewModel: SignupViewModelType
```
FranDepascuali: Why is this var?

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
         bindPasswordElements()
         bindButtons()
         setTextfieldOrder()
+        
+        _viewModel.signUpExecuting.observeNext { [unowned self] executing in
+            if executing {
```
FranDepascuali: can't we do 
```swift
executing ? self._delegate.signupControllerWillExecuteSignUp(self) : self._delegate.signupControllerDidExecuteSignUp(self)
```?

File: Authentication/SignUp/View/SignupController.swift [hunk 10]
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
...
+        _viewModel.signUpErrors.observeNext { [unowned self] in self._delegate.signupController(self, didSignUpWithError: $0) }
     }
     
     private func bindNameElements() {
         if let nameTextField = signupView.usernameTextField {
             _viewModel.name <~ nameTextField.rex_textSignal
             _viewModel.nameValidationErrors.signal.observeNext { [unowned self] errors in
-                self._delegate.signupController(self, didFailNameValidationWithErrors: errors)
+                if errors.isEmpty {
+                    self._delegate.signupControllerDidPassNameValidation(self)
```
FranDepascuali: I think we should change this long naming (brought from objc conventions) and have it written as `didPassNameValidation(self)`

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
...
+                } else {
+                    self._delegate.signupController(self, didFailNameValidationWithErrors: errors)
```
FranDepascuali: I think we should change this long naming (brought from objc conventions) and have it written as `didFailNameValidation(with: )`.

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
...
+                }
             }
             if let nameValidationMessageLabel = signupView.usernameValidationMessageLabel {
                 nameValidationMessageLabel.rex_text <~ _viewModel.nameValidationErrors.signal.map { $0.first ?? " " }
```
FranDepascuali: I think we should express this in a different way, as it is (<~ under the hood do this) observing a new signal every time nameValidationErrors emit a value (and we try to avoid it)

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
...
+            nameTextField.delegate = self
         }
     }
     
     private func bindEmailElements() {
         _viewModel.email <~ signupView.emailTextField.rex_textSignal
         _viewModel.emailValidationErrors.signal.observeNext { [unowned self] errors in
-            self._delegate.signupController(self, didFailEmailValidationWithErrors: errors)
+            if errors.isEmpty {
+                self._delegate.signupControllerDidPassEmailValidation(self)
```
FranDepascuali: I think we should change this long naming (brought from objc conventions) and have it written as didPassEmailValidation(self).

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
...
+            } else {
+                self._delegate.signupController(self, didFailEmailValidationWithErrors: errors)
```
FranDepascuali: I think we should change this long naming (brought from objc conventions) and have it written as didFailEmailValidation(with: ).

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
...
+            }
         }
         if let emailValidationMessageLabel = signupView.emailValidationMessageLabel {
             emailValidationMessageLabel.rex_text <~ _viewModel.emailValidationErrors.signal.map { $0.first ?? " " }
```
FranDepascuali: I think we should express this in a different way, as it is (<~ under the hood do this) observing a new signal every time nameValidationErrors emit a value (and we try to avoid it)

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
...
     }
     
     private func bindPasswordElements() {
-        _viewModel.password <~ signupView.passwordTextField.rex_textSignal
+        _viewModel.password <~ signupView.passwordTextField.rex_textSignal.on(next: { [unowned self] text in
+            if text.isEmpty {
+                self.signupView.passwordVisibilityButton?.hidden = true
+            } else {
+                self.signupView.passwordVisibilityButton?.hidden = false
+            }
```
FranDepascuali: you can write this if as
```swift
    self.signupView.passwordVisibilityButton?.hidden = text.isEmpty
```

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
...
+        })
         _viewModel.passwordValidationErrors.signal.observeNext { [unowned self] errors in
-            self._delegate.signupController(self, didFailPasswordValidationWithErrors: errors)
+            if errors.isEmpty {
+                self._delegate.signupControllerDidPassPasswordValidation(self)
```
FranDepascuali: I think we should change this long naming (brought from objc conventions) and have it written as didPassPasswordValidation(self).

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
...
+            } else {
+                self._delegate.signupController(self, didFailPasswordValidationWithErrors: errors)
```
FranDepascuali: I think we should change this long naming (brought from objc conventions) and have it written as didFailPasswordValidation(with: ).

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
...
+            }
         }
         if let passwordValidationMessageLabel = signupView.passwordValidationMessageLabel {
             passwordValidationMessageLabel.rex_text <~ _viewModel.passwordValidationErrors.signal.map { $0.first ?? " " }
```
FranDepascuali: @@ -6,14 +6,14 @@
  //  Copyright © 2016 Wolox. All rights reserved.		  //  Copyright © 2016 Wolox. All rights reserved.
  //		  //
  		  
 -import Foundation		
  import ReactiveCocoa		  import ReactiveCocoa
  		  
  public final class SignupController: UIViewController {		  public final class SignupController: UIViewController {
      		      
 -    private let _viewModel: SignupViewModelType		 +    private var _viewModel: SignupViewModelType
@FranDepascuali     Wolox member
FranDepascuali added a note 10 minutes ago
Why is this var?
 Add a line note
      private let _signupViewFactory: () -> SignupViewType		      private let _signupViewFactory: () -> SignupViewType
      private let _delegate: SignupControllerDelegate		      private let _delegate: SignupControllerDelegate
 +    private let _transitionDelegate: SignupControllerTransitionDelegate
      		      
      public lazy var signupView: SignupViewType = self._signupViewFactory()		      public lazy var signupView: SignupViewType = self._signupViewFactory()
      		      
 @@ -27,10 +27,11 @@ public final class SignupController: UIViewController {
      // you should not override the `createSignupController` method, but all the others 		      // you should not override the `createSignupController` method, but all the others 
      // that provide the elements this controller uses. (That is to say,		      // that provide the elements this controller uses. (That is to say,
      // `createSignupView`, `createSignupViewModel`, `createSignupControllerDelegate`)		      // `createSignupView`, `createSignupViewModel`, `createSignupControllerDelegate`)
 -    internal init(viewModel: SignupViewModelType, signupViewFactory: () -> SignupViewType, delegate: SignupControllerDelegate = DefaultSignupControllerDelegate()) {		 +    internal init(configuration: SignupControllerConfiguration) {
 -        _viewModel = viewModel		 +        _viewModel = configuration.viewModel
 -        _signupViewFactory = signupViewFactory		 +        _signupViewFactory = configuration.viewFactory
 -        _delegate = delegate		 +        _delegate = configuration.delegate
 +        _transitionDelegate = configuration.transitionDelegate
          super.init(nibName: nil, bundle: nil)		          super.init(nibName: nil, bundle: nil)
          addKeyboardObservers()		          addKeyboardObservers()
      }		      }
 @@ -49,6 +50,7 @@ public final class SignupController: UIViewController {
      		      
      		      
      public override func viewDidLoad() {		      public override func viewDidLoad() {
 +        super.viewDidLoad()
          signupView.render()		          signupView.render()
          bindViewModel()		          bindViewModel()
          let tapRecognizer = UITapGestureRecognizer(target: self, action: #selector(dismissKeyboard))		          let tapRecognizer = UITapGestureRecognizer(target: self, action: #selector(dismissKeyboard))
 @@ -65,62 +67,120 @@ private extension SignupController {
          bindPasswordElements()		          bindPasswordElements()
          bindButtons()		          bindButtons()
          setTextfieldOrder()		          setTextfieldOrder()
 +        
 +        _viewModel.signUpExecuting.observeNext { [unowned self] executing in
 +            if executing {
@FranDepascuali     Wolox member
FranDepascuali added a note 9 minutes ago
can't we do

executing ? self._delegate.signupControllerWillExecuteSignUp(self) : self._delegate.signupControllerDidExecuteSignUp(self)
```?
 Add a line note
 +                self._delegate.signupControllerWillExecuteSignUp(self)
 +            } else {
 +                self._delegate.signupControllerDidExecuteSignUp(self)
 +            }
 +            self.signupView.signUpButtonPressed = executing
 +        }
 +        
 +        _viewModel.signUpErrors.observeNext { [unowned self] in self._delegate.signupController(self, didSignUpWithError: $0) }
      }		      }
      		      
      private func bindNameElements() {		      private func bindNameElements() {
          if let nameTextField = signupView.usernameTextField {		          if let nameTextField = signupView.usernameTextField {
              _viewModel.name <~ nameTextField.rex_textSignal		              _viewModel.name <~ nameTextField.rex_textSignal
              _viewModel.nameValidationErrors.signal.observeNext { [unowned self] errors in		              _viewModel.nameValidationErrors.signal.observeNext { [unowned self] errors in
 -                self._delegate.signupController(self, didFailNameValidationWithErrors: errors)		 +                if errors.isEmpty {
 +                    self._delegate.signupControllerDidPassNameValidation(self)
@FranDepascuali     Wolox member
FranDepascuali added a note 7 minutes ago
I think we should change this long naming (brought from objc conventions) and have it written as didPassNameValidation(self)
 Add a line note
 +                } else {
 +                    self._delegate.signupController(self, didFailNameValidationWithErrors: errors)
@FranDepascuali     Wolox member
FranDepascuali added a note 6 minutes ago
I think we should change this long naming (brought from objc conventions) and have it written as didFailNameValidation(with: ).
 Add a line note
 +                }
              }		              }
              if let nameValidationMessageLabel = signupView.usernameValidationMessageLabel {		              if let nameValidationMessageLabel = signupView.usernameValidationMessageLabel {
                  nameValidationMessageLabel.rex_text <~ _viewModel.nameValidationErrors.signal.map { $0.first ?? " " }		                  nameValidationMessageLabel.rex_text <~ _viewModel.nameValidationErrors.signal.map { $0.first ?? " " }
@FranDepascuali     Wolox member
FranDepascuali added a note 4 minutes ago
I think we should express this in a different way, as it is (<~ under the hood do this) observing a new signal every time nameValidationErrors emit a value (and we try to avoid it)
 Add a line note
              }		              }
 +            nameTextField.delegate = self
          }		          }
      }		      }
      		      
      private func bindEmailElements() {		      private func bindEmailElements() {
          _viewModel.email <~ signupView.emailTextField.rex_textSignal		          _viewModel.email <~ signupView.emailTextField.rex_textSignal
          _viewModel.emailValidationErrors.signal.observeNext { [unowned self] errors in		          _viewModel.emailValidationErrors.signal.observeNext { [unowned self] errors in
 -            self._delegate.signupController(self, didFailEmailValidationWithErrors: errors)		 +            if errors.isEmpty {
 +                self._delegate.signupControllerDidPassEmailValidation(self)
@FranDepascuali     Wolox member
FranDepascuali added a note 3 minutes ago
I think we should change this long naming (brought from objc conventions) and have it written as didPassEmailValidation(self).
 Add a line note
 +            } else {
 +                self._delegate.signupController(self, didFailEmailValidationWithErrors: errors)
@FranDepascuali     Wolox member
FranDepascuali added a note 2 minutes ago
I think we should change this long naming (brought from objc conventions) and have it written as didFailEmailValidation(with: ).
 Add a line note
 +            }
          }		          }
          if let emailValidationMessageLabel = signupView.emailValidationMessageLabel {		          if let emailValidationMessageLabel = signupView.emailValidationMessageLabel {
              emailValidationMessageLabel.rex_text <~ _viewModel.emailValidationErrors.signal.map { $0.first ?? " " }		              emailValidationMessageLabel.rex_text <~ _viewModel.emailValidationErrors.signal.map { $0.first ?? " " }
@FranDepascuali     Wolox member
FranDepascuali added a note 2 minutes ago
I think we should express this in a different way, as it is (<~ under the hood do this) observing a new signal every time nameValidationErrors emit a value (and we try to avoid it)

Comment (FranDepascuali) [inline Authentication/SignUp/View/SignupController.swift]: I think we should express this in a different way, as it is (<~ under the hood do this) observing a new signal every time nameValidationErrors emit a value (and we try to avoid it)

File: Authentication/SignUp/View/SignupController.swift [hunk 20]
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
...
+    }
+    
+    func bindPasswordConfirmationElements() {
+        if let passwordConfirmationTextField = signupView.passwordConfirmTextField {
+            _viewModel.passwordConfirmation <~ passwordConfirmationTextField.rex_textSignal.on(next: { [unowned self] text in
+                if text.isEmpty {
+                    self.signupView.passwordConfirmVisibilityButton?.hidden = true
+                } else {
+                    self.signupView.passwordConfirmVisibilityButton?.hidden = false
+                }
```
FranDepascuali: you can do this as 
```swift
self.signupView.passwordConfirmVisibilityButton?.hidden = text.isEmpty
```

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
...
+                })
+            _viewModel.passwordConfirmationValidationErrors.signal.observeNext { [unowned self] errors in
+                if errors.isEmpty {
+                    self._delegate.signupControllerDidPassPasswordConfirmationValidation(self)
```
FranDepascuali: We should rename this function

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
...
+                } else {
+                    self._delegate.signupController(self, didFailPasswordConfirmationValidationWithErrors: errors)
```
FranDepascuali: We should rename this function

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,120 @@ private extension SignupController {
...
+                }
+            }
+            if let passwordConfirmValidationMessageLabel = signupView.passwordConfirmValidationMessageLabel {
+                passwordConfirmValidationMessageLabel.rex_text <~ _viewModel.passwordConfirmationValidationErrors.signal.map { $0.first ?? " " }
```
FranDepascuali: We should have in the future a discussion of whether its okay to only show the first error. (we can leave it like this right now)

Authentication/SignUp/View/SignupControllerDelegate.swift
```diff
@@ -41,7 +41,7 @@ extension SignupControllerDelegate {
     public var shouldDisplaySignupErrorWithAlert: Bool { return true }
     
     public func signupControllerWillExecuteSignUp(controller: SignupController) {
-        if let errorLabel = controller.signupView.signupErrorLabel {
+        if let errorLabel = controller.signupView.signUpErrorLabel {
             errorLabel.text = " "
```
FranDepascuali: Why are we doing this here?

Authentication/SignUp/View/SignupControllerDelegate.swift
```diff
@@ -41,7 +41,7 @@ extension SignupControllerDelegate {
...
         }
         let app = UIApplication.sharedApplication()
```
FranDepascuali: I think this is okay, though it somehow makes some noise to me

Authentication/SignUp/View/SignupControllerDelegate.swift
```diff
@@ -53,19 +53,13 @@ extension SignupControllerDelegate {
...
-        controller.signupView.passwordTextFieldValid = false
-        controller.signupView.passwordConfirmationTextFieldValid = false
-        controller.signupView.signupButtonPressed = false
-        
-        if let errorLabel = controller.signupView.signupErrorLabel {
+    public func signupController(controller: SignupController, didSignUpWithError error: SessionServiceError) {        
+        if let errorLabel = controller.signupView.signUpErrorLabel {
             errorLabel.text = error.message
         }
         if shouldDisplaySignupErrorWithAlert {
```
FranDepascuali: I would rename this to `shouldDisplayError`

Authentication/SignUp/View/SignupView.swift
```diff
@@ -6,101 +6,370 @@
...
+    @IBOutlet weak var loginLabelOutlet: UILabel! {
+        didSet { loginLabelOutlet.text = loginLabelText }
+    }
+    
+    internal var loginButton: UIButton { return loginButtonOutlet }
+    @IBOutlet weak var loginButtonOutlet: UIButton! {
+        didSet { loginButtonOutlet.setTitle(loginButtonTitle, forState: .Normal) }
+    }
+    
+    internal var usernameTextFieldValid: Bool = false { didSet { usernameTextFieldValidWasSet() } }
```
FranDepascuali: Delete `: Bool`

Authentication/SignUp/View/SignupView.swift
```diff
@@ -6,101 +6,370 @@
...
+    internal var passwordTextFieldValid: Bool = false { didSet { passwordTextFieldValidWasSet() } }
+    internal var passwordTextFieldSelected: Bool = false { didSet { passwordTextFieldSelectedWasSet() } }
+    internal var showPassword: Bool = false { didSet { showPasswordWasSet() } }
+    
+    internal var passwordConfirmationTextFieldValid: Bool = false { didSet { passwordConfirmationTextFieldValidWasSet() } }
+    internal var passwordConfirmationTextFieldSelected: Bool = false { didSet { passwordConfirmationTextFieldSelectedWasSet() } }
+    internal var showConfirmationPassword: Bool = false { didSet { showConfirmationPasswordWasSet() } }
+    
+    internal var signUpButtonEnabled: Bool = false { didSet { signUpButtonEnabledWasSet() } }
+    internal var signUpButtonPressed: Bool = false { didSet { signUpButtonPressedWasSet() } }
```
FranDepascuali: Delete `: Bool` in all these properties

Authentication/SignUp/View/SignupView.swift
```diff
@@ -6,101 +6,370 @@
...
+                color = delegate.colorPalette.textfieldsNormal.CGColor
+                usernameErrorsHeightConstraint?.constant = 0
+                usernameErrorsHeightConstraint?.active = true
+            } else {
+                color = delegate.colorPalette.textfieldsError.CGColor
+                usernameErrorsHeightConstraint?.active = false
+            }
+            usernameTextField?.layer.borderColor = color
+        } else {
+            usernameErrorsHeightConstraint?.constant = 0
```
FranDepascuali: Wasn't it easier to use `UIStackView` or you couldn't?

Authentication/SignUp/ViewModel/SignupViewModel.swift
```diff
@@ -22,9 +22,13 @@ public protocol SignupViewModelType {
     
     var password: MutableProperty<String> { get }
     var passwordValidationErrors: AnyProperty<[String]> { get }
+    var showPassword: MutableProperty<Bool> { get }
```
FranDepascuali: I would rename this to
```swift
passwordVisible
```

Authentication/SignUp/ViewModel/SignupViewModel.swift
```diff
@@ -22,9 +22,13 @@ public protocol SignupViewModelType {
...
+    var togglePasswordVisibilityCocoaAction: CocoaAction { get }
     
     var passwordConfirmation: MutableProperty<String> { get }
     var passwordConfirmationValidationErrors: AnyProperty<[String]> { get }
+    var showConfirmationPassword: MutableProperty<Bool> { get }
```
FranDepascuali: I would rename this to
```swift
 var confirmationPasswordVisible: MutableProperty<Bool> { get }
```

Authentication/SignUp/ViewModel/SignupViewModel.swift
```diff
@@ -22,9 +22,13 @@ public protocol SignupViewModelType {
...
+    var toggleConfirmPswdVisibilityCocoaAction: CocoaAction { get }
```
FranDepascuali: avoid the pswd abbreviation

Authentication/SignUp/ViewModel/SignupViewModel.swift
```diff
@@ -47,54 +52,85 @@ public final class SignupViewModel<User: UserType, SessionService: SessionServic
...
         passwordConfirmationValidationErrors = AnyProperty(initialValue: [], signal: passwordConfirmValidationResult.map { $0.errors })
         
-        _credentialsAreValid = AnyProperty<Bool>(initialValue: false, signal: nameValidationResult.map { $0.isValid })
-                            .and(AnyProperty<Bool>(initialValue: false, signal: emailValidationResult.map { $0.isValid }))
-                            .and(AnyProperty<Bool>(initialValue: false, signal: passwordValidationResult.map { $0.isValid }))
-                            .and(AnyProperty<Bool>(initialValue: false, signal:passwordConfirmValidationResult.map { $0.isValid }))
+        var credentialsAreValid = AnyProperty<Bool>(initialValue: false, signal: emailValidationResult.map { $0.isValid })
+            .and(AnyProperty<Bool>(initialValue: false, signal: passwordValidationResult.map { $0.isValid }))
+        if usernameEnabled {
+            credentialsAreValid = credentialsAreValid
```
FranDepascuali: why are you doing this?

Authentication/SignUp/ViewModel/SignupViewModel.swift
```diff
@@ -47,54 +52,85 @@ public final class SignupViewModel<User: UserType, SessionService: SessionServic
...
+    private lazy var _togglePasswordVisibility: Action<AnyObject, Bool, NoError> = self.initializeTogglePasswordVisibilityAction()
+    private lazy var _toggleConfirmationPasswordVisibility: Action<AnyObject, Bool, NoError> = self.initializeToggleConfirmationPasswordVisibilityAction()
+    
+    internal init(sessionService: SessionService,
+         credentialsValidator: SignupCredentialsValidator = SignupCredentialsValidator(),
+         passwordConfirmationEnabled: Bool = true,
+         usernameEnabled: Bool = true) {
         _sessionService = sessionService
         
         let nameValidationResult = name.signal.map(credentialsValidator.nameValidator.validate)
```
FranDepascuali: I would avoid naming this Result, one can think that it is talking about Result dependency

Authentication/SignUp/ViewModel/SignupViewModel.swift
```diff
@@ -47,54 +52,85 @@ public final class SignupViewModel<User: UserType, SessionService: SessionServic
...
         let emailValidationResult = email.signal.map(credentialsValidator.emailValidator.validate)
```
FranDepascuali: I would avoid naming this Result, one can think that it is talking about Result dependency

Authentication/SignUp/ViewModel/SignupViewModel.swift
```diff
@@ -47,54 +52,85 @@ public final class SignupViewModel<User: UserType, SessionService: SessionServic
...
         let passwordValidationResult = password.signal.map(credentialsValidator.passwordValidator.validate)
```
FranDepascuali: I would avoid naming this Result, one can think that it is talking about Result dependency

Authentication/SignUp/ViewModel/SignupViewModel.swift
```diff
@@ -47,54 +52,85 @@ public final class SignupViewModel<User: UserType, SessionService: SessionServic
...
         let passwordConfirmValidationResult = combineLatest(password.signal, passwordConfirmation.signal)
```
FranDepascuali: I would avoid naming this Result, one can think that it is talking about Result dependency

Authentication/SignUp/ViewModel/SignupViewModel.swift
```diff
@@ -47,54 +52,85 @@ public final class SignupViewModel<User: UserType, SessionService: SessionServic
...
+            self.showConfirmationPassword.value = !self.showConfirmationPassword.value
+            return SignalProducer(value: self.showConfirmationPassword.value).observeOn(UIScheduler())
+        }
+    }
+    
 }
 
-private func getTermsAndServicesValidationResultFromAcceptance(accepted: Bool) -> ValidationResult {
-    if accepted {
+private func getPasswordConfirmValidationResultFromEquality(equals: Bool) -> ValidationResult {
```
FranDepascuali:
```swift
private func passwordsMatch(equals: Bool) {
    return equals ? .Valid : .Invalid("signup-error.password-confirmation.invalid".localized)
}
```

Authentication/Login/View/LoginView.swift
danielaRiesgo: The thing is the file was very long and swiftlint complained. I know it's not the best solution, but it was a little difference of lines.

Authentication/SignUp/View/SignupController.swift
danielaRiesgo: Sorry, that remained from a time where the controller set a property of the view model.

Authentication/SignUp/View/SignupController.swift
danielaRiesgo: Yes. For now what we said is here we show the first error in the label, and then in the delegate if they want they can do something else with the errors, like replacing the label.

Authentication/SignUp/View/SignupControllerDelegate.swift
danielaRiesgo: For it to clean previous errors of signup, so that they can be distinguished from the possible new ones.

Authentication/SignUp/View/SignupControllerDelegate.swift
danielaRiesgo: I wouldn't do that because there are many possible validation errors, but this refers to only sign up errors, and it's not representing displaying the error in any form, but with an alert. Because in the label it will always display it.

Authentication/SignUp/ViewModel/SignupViewModel.swift
danielaRiesgo: But it's type is `ValidationResult`

Authentication/SignUp/ViewModel/SignupViewModel.swift
danielaRiesgo: Because if there is going to be a username textfield, validation should include that field.

Authentication/SignUp/View/SignupController.swift
guidomb: I don't like that. I also used to use the conditional operator for this sort of things but now I think that this operator shouldn't be used for control flow. Calling a delegate could potentially imply a side effect. I would prefer to make it more explicit by using `if else`.

I would suggest to only use `?:` when the things that you are returning are values and don't perform side effects. Also they should be short. If the line is too long is better to use `if/else`. Less lines doesn't necessary mean better readability,

Authentication/SignUp/View/SignupController.swift
guidomb: Yeah originally we decided that the default implementation should only show the first error for the sake of simplicity

Authentication/SignUp/View/SignupControllerDelegate.swift
guidomb: I think @FranDepascuali meant why are we changing from `signup` to `signUp`. I think it should be `signup` because it should be used as a noun.

Authentication/SignUp/View/SignupControllerDelegate.swift
guidomb: Shouldn't be used as a noun? I such case it should be `signup`.

Authentication/SignUp/View/SignupControllerDelegate.swift
guidomb: I agree with @danielaRiesgo in this one.

Authentication/SignUp/ViewModel/SignupViewModel.swift
guidomb: Why? Is a matter of taste or you have a stronger argument?

guidomb: Now that I think about a little bit, using verb is better for method than properties. Which makes `passwordVisible` a better candidate.

Authentication/SignUp/ViewModel/SignupViewModel.swift
guidomb: Yeah I also don't like when we use type names in variable names but sometimes you can't think of a better name. Maybe `isNameValid`?

guidomb:
Sorry guys, I don't want to add noise to the conversation but now I realized that @danielaRiesgo applied one of the suggestion made by @FranDepascuali Which I don't agree with. Nothing critical but I think it would be nice to have a conversation about it and maybe define a rule for future reference.

I am talking about [this](https://github.com/Wolox/wolmo-authentication-ios/pull/10/commits/f618c9de9eb21cbe4a83d701c99ba9a2699e7329) commit.

Authentication/SignUp/View/SignupControllerDelegate.swift
danielaRiesgo: Ah, sorry. I changed it so to be easier to associate with the sign up button which does the action, since the label is for that action's error, not the whole signup page.
Do I change it back?

Authentication/SignUp/ViewModel/SignupViewModel.swift
danielaRiesgo: Yes. I liked the new name, because `showPassword` may lead to confusion like true because we should make it show or because it is already shown", etc.

danielaRiesgo:
Yes, I agree with @guidomb to turn it back to `if-else` because of making side effects evident.

Maybe we can establish that I don't correct anything until both have reviewed the new commits. What do you say?

Authentication/SignUp/ViewModel/SignupViewModel.swift
danielaRiesgo: I wouldn't like that because it's not a `true-false` signal like in `_credentialsAreValid`. Maybe `nameValidation` alone isn't so bad.

Authentication/SignUp/View/SignupControllerDelegate.swift
guidomb: Personally I think it should be `signup` but I made this mistake so many times too.

Authentication/SignUp/ViewModel/SignupViewModel.swift
guidomb: I that case I think `validationResult` is clearer.

guidomb:
@danielaRiesgo Yeah but I don't want to be the blocker. Most probably you two are going to be faster than me reviewing this sort of PR. Every now and them we may have different opinions of certain thinks and this "double" work could happen. The only thing (I can think of) to avoid this is that if somebody makes a correction that you are not too sure about, name another person to join the conversation. With time we should eventually settle all conventions.

Authentication/SignUp/View/SignupController.swift
FranDepascuali: @guidomb I don't agree in this one. They both are a way of expressing conditional logic; and their functionality is the same. I find no reason for which you would use `if let` rather than the operator because it implies side effects.

I think that when decision is between two elements, one should try to use the operator. It's more accurate in representing what we want to do, it's clearer to the reader. One do not think about things with the `if/let` mentality; one decides between two options in the same place. 

I agree with you in the length of the function.

I found this phrase that somehow sums up what I tried to say:
```
An if/else statement emphasises the branching first and what's to be done is secondary, while a ternary operator emphasises what's to be done over the selection of the values to do it with.
http://stackoverflow.com/questions/4192225/ternary-operator-vs-if-statement-question-of-prettiness

```

FranDepascuali:
I think that when decision is between two elements, one should try to use the operator. It's more accurate in representing what we want to do, it's clearer to the reader. One do not think about things with the `if/let` mentality; one decides between two options in the same place. 

I agree with you in the length of the function.

I found this phrase that somehow sums up what I tried to say:
```
An if/else statement emphasises the branching first and what's to be done is secondary, while a ternary operator emphasises what's to be done over the selection of the values to do it with.
```
http://stackoverflow.com/questions/4192225/ternary-operator-vs-if-statement-question-of-prettiness

Why is side-effect involved here?

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,102 @@ private extension SignupController {
...
+    
+    func bindPasswordConfirmationElements() {
+        if let passwordConfirmationTextField = signupView.passwordConfirmTextField {
+            _viewModel.passwordConfirmation <~ passwordConfirmationTextField.rex_textSignal.on(next: { [unowned self] text in
+                self.signupView.passwordConfirmVisibilityButton?.hidden = text.isEmpty
+            })
+            _viewModel.passwordConfirmationValidationErrors.signal.observeNext { [unowned self] errors in
+                errors.isEmpty
+                    ? self._delegate.didPassPasswordConfirmationValidation(self)
+                    : self._delegate.didFailPasswordConfirmationValidation(self, with: errors)
```
FranDepascuali: Sorry because I corrected this, you should change it back to `if else`

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,102 @@ private extension SignupController {
...
     private func bindPasswordElements() {
-        _viewModel.password <~ signupView.passwordTextField.rex_textSignal
+        _viewModel.password <~ signupView.passwordTextField.rex_textSignal.on(next: { [unowned self] text in
+            self.signupView.passwordVisibilityButton?.hidden = text.isEmpty
+        })
         _viewModel.passwordValidationErrors.signal.observeNext { [unowned self] errors in
-            self._delegate.signupController(self, didFailPasswordValidationWithErrors: errors)
+            errors.isEmpty
+                ? self._delegate.didPassPasswordValidation(self)
+                : self._delegate.didFailPasswordValidation(self, with: errors)
```
FranDepascuali: Sorry because I corrected this, you should change it back to `if else`

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,102 @@ private extension SignupController {
...
+        _viewModel.signUpErrors.observeNext { [unowned self] in self._delegate.didSignUp(self, with: $0) }
     }
     
     private func bindNameElements() {
         if let nameTextField = signupView.usernameTextField {
             _viewModel.name <~ nameTextField.rex_textSignal
             _viewModel.nameValidationErrors.signal.observeNext { [unowned self] errors in
-                self._delegate.signupController(self, didFailNameValidationWithErrors: errors)
+                errors.isEmpty
+                    ? self._delegate.didPassNameValidation(self)
```
FranDepascuali: Sorry because I corrected this, you should change it back to `if else`

Authentication/SignUp/View/SignupController.swift
```diff
@@ -65,62 +67,102 @@ private extension SignupController {
         bindPasswordElements()
         bindButtons()
         setTextfieldOrder()
+        
+        _viewModel.signUpExecuting.observeNext { [unowned self] executing in
+            executing
+                ? self._delegate.willExecuteSignUp(self)
+                : self._delegate.didExecuteSignUp(self)
```
FranDepascuali: Sorry because I corrected this, you should change it back to `if else`

Authentication/Login/View/LoginController.swift
```diff
@@ -89,25 +90,21 @@ private extension LoginController {
...
         
         _viewModel.logInExecuting.observeNext { [unowned self] executing in
-            if executing {
-                self._delegate.loginControllerWillExecuteLogIn(self)
-            } else {
-                self._delegate.loginControllerDidExecuteLogIn(self)
-            }
+            executing
+                ? self._delegate.willExecuteLogIn(self)
+                : self._delegate.didExecuteLogIn(self)
```
FranDepascuali: Sorry because I corrected this, you should change it back to `if else`

Authentication/Login/View/LoginController.swift
```diff
@@ -89,25 +90,21 @@ private extension LoginController {
...
         _viewModel.email <~ loginView.emailTextField.rex_textSignal
         _viewModel.emailValidationErrors.signal.observeNext { [unowned self] errors in
-            if errors.isEmpty {
-                self._delegate.loginControllerDidPassEmailValidation(self)
-            } else {
-                self._delegate.loginController(self, didFailEmailValidationWithErrors: errors)
-            }
+            errors.isEmpty
+                ? self._delegate.didPassEmailValidation(self)
+                : self._delegate.didFailEmailValidation(self, with: errors)
```
FranDepascuali: Sorry because I corrected this, you should change it back to `if else`

Authentication/Login/View/LoginController.swift
```diff
@@ -117,26 +114,20 @@ private extension LoginController {
...
         })
         _viewModel.passwordValidationErrors.signal.observeNext { [unowned self] errors in
-            if errors.isEmpty {
-                self._delegate.loginControllerDidPassPasswordValidation(self)
-            } else {
-                self._delegate.loginController(self, didFailPasswordValidationWithErrors: errors)
-            }
+            errors.isEmpty
+                ? self._delegate.didPassPasswordValidation(self)
+                : self._delegate.didFailPasswordValidation(self, with: errors)
```
FranDepascuali: Sorry because I corrected this, you should change it back to `if else`

FranDepascuali:
correct that and I'm 👍

Authentication/SignUp/View/SignupController.swift
danielaRiesgo: No problem !

woloxci:

script/.env
```diff
@@ -11,4 +12,4 @@ IOS_DESTINATION_VERSION=${IOS_DESTINATION_VERSION:-"latest"}
 IOS_DESTINATION_SIMULATOR_NAME=${IOS_DESTINATION_SIMULATOR_NAME:-"iPhone 6"}
 OSX_DESTINATION_ARCH=${OSX_DESTINATION_ARCH:-""}
 INSTALL_GITHOOKS=true
-REQUIRED_XCODE_VERSION=7.3.1
+FORCE_SWIFTLINT_VERSION=true
```
guidomb: I would suggest to put this when you are calling the `script/cibuild` instead of putting it here because you do want to force the CI machine to force swiftlint update but you may not want to do that on a dev machine. The developer should be the one that explicitly decides to update the swiftlint version.

danielaRiesgo: Great. The same happened to me with carthage in git hooks and it was annoying because I had two projects with different carthage versions. Maybe we could change that too?

guidomb:
👍

woloxci:

PR #11, opened 2016-07-19
Adding core and fixing localization

danielaRiesgo:
## Sumary ##
**Trello card**: https://trello.com/c/16w2TUeF/27-incluir-wolmo-core-como-dependencia
**Trello card**: https://trello.com/c/bQyWymgh/23-lograr-string-localization-automatica

Adding wolmo-core-ios as dependency.
Making localization work in two steps: check for localized string in main bundle, if not check in framework bundle.

### Localizing just the `New to AuthenticationDemo?`
simulator screen shot jul 25 2016 3 21 33 pm


woloxci:

Authentication/AuthenticationBootstrapper.swift
```diff
@@ -162,7 +162,7 @@ public class AuthenticationBootstrapper<User: UserType, SessionService: SessionS
     }
     
     public func createSignupView() -> SignupViewType {
-        let view = SignupView.loadFromNib()
+        let view = SignupView.loadFromNib()!
```
guidomb: I think we should handle the nil case here and raise an exception and log an error explaining what the problem might be and how it should be solved. Either that or make this method return an optional

danielaRiesgo: I don't think this method should return an optional since a view must be provided for the authentication process.
About the exception, we could, but I don't think it's necessary since this view comes with the framework, it will never fail.

guidomb: OK

guidomb:
👍

woloxci:

guidomb:
👍

Authentication/Extensions.swift
```diff
@@ -2,20 +2,42 @@
...
 //  Copyright © 2016 Wolox. All rights reserved.
 //
 
 import Foundation
 
 public extension String {
     
-    var localized: String {
-        return NSLocalizedString(self, tableName: nil, bundle: NSBundle.mainBundle(), value: "", comment: "")
+    public var frameworkBundle: NSBundle { return  NSBundle(forClass: LoginView.self) }
```
FranDepascuali: I think this should be internal.

Authentication/Extensions.swift
```diff
@@ -2,20 +2,42 @@
...
+    
+    /**
+         Returns a localized representation of the string,
+         searching first in the main bundle or if it does
+         not have the key in the framework one if.
+     */
+    public var frameworkLocalized: String {
```
FranDepascuali: I think this should be internal.

Authentication/Extensions.swift
```diff
@@ -2,20 +2,42 @@
...
     }
     
-    func localized(arguments: CVarArgType...) -> String {
-        return String(format: NSLocalizedString(self, tableName: nil, bundle: NSBundle.mainBundle(), value: "", comment: ""), arguments: arguments)
+    /**
+         Returns a localized representation of the string.
+         
+         - parameter arguments: Formatting arguments.
+     */
+    public func frameworkLocalized(arguments: CVarArgType...) -> String {
```
FranDepascuali: I think this should be internal.

Authentication/Extensions.swift
FranDepascuali: Also, I don't think frameworkBundle should be a var in an extension of String. It should be a free let.

Authentication/AuthenticationBootstrapper.swift
```diff
@@ -31,6 +31,8 @@ public class AuthenticationBootstrapper<User: UserType, SessionService: SessionS
         return sessionService.currentUser.value
     }
 
+    private let frameworkBundle: NSBundle = NSBundle(forClass: LoginView.self)
```
FranDepascuali: I don't like having this in all the classes which have a view. I would rather have (in a separate, common file) a 
```
    internal let FrameworkBundle = NSBundle(forClass: LoginView.self)
```
and then use that one.

Authentication/Extensions.swift
FranDepascuali: what do you think of this?
```swift
    public var frameworkLocalized: String {
        let localized: (from: NSBundle) -> String = {
            NSLocalizedString(self, tableName: nil, bundle: $0, value: "", comment: "")
        }
        
        let mainLocalized = localized(from: NSBundle.mainBundle())
        return mainLocalized == self ? localized(from: frameworkBundle) : mainLocalized
    }
```

Authentication/Extensions.swift
FranDepascuali: I think you could use the other localized function:
```swift
    public func frameworkLocalized(arguments: CVarArgType...) -> String {
        return String(format: frameworkLocalized, arguments: arguments)
    }
```

Authentication/Extensions.swift
FranDepascuali: Also, I think this should be named localized, rather than `frameworkLocalized`. Frameworklocalized make me think that it searches the localization just from the framework, but as you did (very good) it looks for it in main bundle first and then in the framework, so it should be just localized. What do you think? @danielaRiesgo @guidomb

danielaRiesgo: Wouldn't there be conflicts with Core's localized?

FranDepascuali:
👍

woloxci:

PR #15, opened 2016-07-26
Terms and services

danielaRiesgo:
## Sumary ##
**Trello card**: https://trello.com/c/PmwSAhKB/18-implementar-terms-and-services

Adding terms and services functionality.
Adding controller with web view.
Adding obligatory URL in sign up configuration.
Adding to view a function to configure textview when having the url.
Solving file_length linter violations in `SignupView`, not in `AuthenticationBootstrapper`.

## Screencast ##
(Using any example page)
terms and services


guidomb:
👍

Authentication/AuthenticationBootstrapper.swift
```diff
@@ -63,6 +62,13 @@ public class AuthenticationBootstrapper<User: UserType, SessionService: SessionS
             }
         }
     }
+    
+    public convenience init(sessionService: SessionService, window: UIWindow, termsAndServicesURL: NSURL,
+                            mainViewControllerFactory: () -> UIViewController) {
+        self.init(sessionService: sessionService, window: window,
+                  viewConfiguration: AuthenticationViewConfiguration(termsAndServicesURL: termsAndServicesURL),
+                  mainViewControllerFactory: mainViewControllerFactory)
+    }
```
FranDepascuali: Why was this changed?

Authentication/SignUp/TermsAndServices/TermsAndServicesController.swift
```diff
@@ -0,0 +1,67 @@
...
+    
+    public override func loadView() {
+        let preferences = WKPreferences()
+        preferences.javaScriptEnabled = true
+        preferences.minimumFontSize = 7
+        
+        let configuration = WKWebViewConfiguration()
+        configuration.preferences = preferences
+        
+        let webView = WKWebView(frame: .zero, configuration: configuration)
```
FranDepascuali: I think this should be done with constraints rather than frames. How does it work? I bet the frame is different in loadView() than viewDidAppear() because it is not yet rendered

Authentication/SignUp/TermsAndServices/TermsAndServicesController.swift
```diff
@@ -0,0 +1,67 @@
...
+        super.viewDidLoad()
+        navigationController?.navigationBarHidden = false
+        let request = NSURLRequest(URL: _url)
+        _webView.loadRequest(request)
+        _delegate.didStartLoadingTermsAndServices(self)
+    }
+    
+    public override func viewWillDisappear(animated: Bool) {
+        super.viewWillDisappear(animated)
+        navigationController?.navigationBarHidden = true
```
FranDepascuali: What happens if you open terms and service, lock the phone (or press home button) and then re-enter it? NavigationBarHidden will remain in true, and I think you want it in false?

Authentication/AuthenticationBootstrapper.swift
danielaRiesgo: Because there is no possible default terms and services URL, so now there is no default `SignupViewConfiguration` and so there is no default `AuthenticationViewConfiguration`. But the most basic `AuthenticationViewConfiguration` receives an url, so I created this convenience init to make it easier for the user to create the bootstrapper with the url without having to create the `AuthenticationViewConfiguration`.

Authentication/SignUp/TermsAndServices/TermsAndServicesController.swift
danielaRiesgo: So I could just move the `navigationController?.navigationBarHidden = false` to the `viewDidAppear` method, right?

Authentication/SignUp/TermsAndServices/TermsAndServicesController.swift
danielaRiesgo: I did it with constraints and it didn't work. Because this view isn't content of any other view, this is the controller's view (`view = webView`), so I guess the constraint to the screen are "automatic".

FranDepascuali:
👍

PR #16, opened 2016-07-27
Making initial screen selectable by user between login and signup

danielaRiesgo:
## Sumary ##
**Trello card**: https://trello.com/c/184vqPn5/37-hacer-que-authentication-pueda-empezar-en-login-o-signup-a-eleccion

Making initial screen selectable by user between login and signup.
And making the navigation in the way: if not initial screen, popViewController.
It didn't include any visual change to views.

### Login first
login first

### Signup first
signup first

### DIFF
--- a/Authentication/AuthenticationBootstrapper.swift
+++ b/Authentication/AuthenticationBootstrapper.swift
@@ -8,6 +8,10 @@
 
 import Core
 
+public enum AuthenticationInitialScreen {
+    case Login, Signup
+}
+
 /**
     Bootstrapper to start the application.
     Takes care of starting the authentication process before the main View Controller of the app when necessary,
@@ -28,7 +32,9 @@ public class AuthenticationBootstrapper<User: UserType, SessionService: SessionS
     /// The configuration that defines colour and fonts and assets, like the logo, used in the views.
     /// It also includes other configurations like the textfields selected to use in signup.
     private let _viewConfiguration: AuthenticationViewConfiguration
-
+    /// Property indicating the authentication screen to be shown the first time.
+    private let _initialScreen: AuthenticationInitialScreen
+    
     
     /**
         Initializes a new authentication bootstrapper with the session service to use for logging in and out and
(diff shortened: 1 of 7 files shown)


woloxci:

guidomb:
👍

Authentication/AuthenticationBootstrapper.swift
```diff
@@ -219,33 +235,29 @@ public class AuthenticationBootstrapper<User: UserType, SessionService: SessionS
 extension AuthenticationBootstrapper: LoginControllerTransitionDelegate {
     
     public func onSignup(controller: LoginController) {
```
FranDepascuali: What does this `onSignup` means? When the user sign up or when he presses the button? If it is the second one, we should change its name

Authentication/AuthenticationBootstrapper.swift
```diff
@@ -219,33 +235,29 @@ public class AuthenticationBootstrapper<User: UserType, SessionService: SessionS
...
-            _window.rootViewController = UINavigationController(rootViewController: recoverPasswordController)
-        }
+        controller.navigationController!.pushViewController(recoverPasswordController, animated: true)
     }
     
 }
 
 extension AuthenticationBootstrapper: SignupControllerTransitionDelegate {
     
     public func onLogin(controller: SignupController) {
```
FranDepascuali: What does this onLogin means? When the user logins or when he presses the button? If it is the second one, we should change its name

Authentication/AuthenticationBootstrapper.swift
danielaRiesgo: When the user is in the login screen and presses the `Sign Up` button. The documentation branch adds this explanation here.

Authentication/AuthenticationBootstrapper.swift
danielaRiesgo: When the user is in the signup screen and presses the `Log In` button. The documentation branch adds this explanation here.

Authentication/AuthenticationBootstrapper.swift
FranDepascuali: I know I told you to change it in other PR, but now I think that it would be more appropiate to name it `onSignupButtonPressed`. We should strive for "clarity on the place of use". Sorry for suggesting this change in thr other PR.

I know that because the function is in the LoginControllerTransitionDelegate, we can infer that it is from that protocol. But that is because we know what it means because we are designing the library. Furthermore, If you implement both delegates in the same extensions, you lose this "reference".
What do you think @guidomb @danielaRiesgo?

guidomb: This one is a tricky one, in a regular app I would tell you that the name should communicate intention, so `onSignup` would be a better choice. But in this particular case I get your point. Maybe calling it `performTransitionToSignup`

danielaRiesgo: I like `onSignupButtonPressed`, but as is the transition delegate it may be good to state in the name of the function that they must do the transition. So maybe `toSignup` or `performTransitionToSignup` are better.

woloxci:

PR #18, opened 2016-07-27
Signup view model tests

danielaRiesgo:
## Sumary ##
**Trello card**: https://trello.com/c/FJ6N5DPv/33-hacer-tests-del-signupviewmodel

Adding sign up view model tests.
Refactoring login view model tests when refactoring session service mock for tests to test view model and not mock logic.
Adding toggle visibility tests to login view model ones.

### DIFF
--- a/AuthenticationTests/LogIn/LoginViewModelTests.swift
+++ b/AuthenticationTests/LogIn/LoginViewModelTests.swift
@@ -18,12 +18,12 @@ class LoginViewModelSpec: QuickSpec {
     
     override func spec() {
         describe("LoginViewModel") {
-            
+
             var sessionService: MockSessionService!
             var loginVM: LoginViewModel<MyUser, MockSessionService>!
             
             beforeEach() {
-                sessionService = MockSessionService(email: Email(raw: "myuser@mail.com")!, password: "password", username: "MyUser")
+                sessionService = MockSessionService()
                 loginVM = LoginViewModel(sessionService: sessionService)
             }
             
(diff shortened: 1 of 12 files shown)


guidomb:
👍

AuthenticationTests/LogIn/LoginViewModelTests.swift
```diff
@@ -18,12 +18,12 @@ class LoginViewModelSpec: QuickSpec {
     
     override func spec() {
         describe("LoginViewModel") {
-            
+
             var sessionService: MockSessionService!
             var loginVM: LoginViewModel<MyUser, MockSessionService>!
```
FranDepascuali: this should be `loginViewModel`

AuthenticationTests/LogIn/LoginViewModelTests.swift
```diff
@@ -36,12 +36,36 @@ class LoginViewModelSpec: QuickSpec {
                 expect(loginVM.passwordVisible.value) == false
             }
             
+            describe("#togglePasswordVisibility") {
+                
+                context("when executing action") {
+                    
+                    it("should change #passwordVisible from false to true") { waitUntil { done in
```
FranDepascuali: I would move the false and true part to the context:
```swift
context("When executing action") {
    context("When #passwordVisible is false") {
        it("should change it to true") {
        }
    }
    context("When #passwordVisible is true") {
        it("should change it to false") {
        }
    }
}
```
However, you decide, it isn't so important so if you want to leave it like this its okay

FranDepascuali:
👍

PR #22, opened 2016-08-02
Bootstrapper without inheritance

danielaRiesgo:
## Sumary ##
**Trello card**: https://trello.com/c/NZES2fw3/38-refactorizar-bootstrapper-para-evitar-herencia

Created a AuthenticationComponentsFactory protocol to avoid having the bootstrapper creating everything.
Adding default implementation for all creation methods possible, but leaving "open" the required methods, like `createMainViewController` and `createXxxxViewConfiguration`. But adding a default struct that conforms to protocol to easily create a protocol-conforming object being able to configure those required elements.
Changing `Bootstrapper` name to `Coordinator`.

## Known issues ##
We should discuss if the session service should go in the ComponentsFactory or not. The issue with having it there is that it will need an associated type because SessionServiceType uses an associated type.


woloxci:

Authentication/AuthenticationComponentsFactory.swift
```diff
@@ -0,0 +1,353 @@
...
+//  AuthenticationComponentsFactory.swift
+//  Authentication
+//
+//  Created by Daniela Riesgo on 8/1/16.
+//  Copyright © 2016 Wolox. All rights reserved.
+//
+
+public protocol AuthenticationComponentsFactoryType: LoginComponentsFactory, SignupComponentsFactory {
+    
+    //var sessionService: SessionServiceType { get }
```
guidomb: ???

Authentication/AuthenticationComponentsFactory.swift
```diff
@@ -0,0 +1,353 @@
...
+             - initialScreen: authentication screen to be shown
+                the first time. By default, Login.
+             - logo: logo to add to the default LoginViewConfiguration.
+             - termsAndServicesURL: terms and services URL used
+                to create the default SignupViewConfiguration.
+             - mainControllerFactory: main view controller factory
+                method to trigger when asked for the main view
+                controller.
+     */
+    public init(initialScreen: AuthenticationInitialScreen = .Login,
```
guidomb: I think this should be a conveniece initializer if that works with structs.

Authentication/AuthenticationCoordinator.swift
```diff
@@ -0,0 +1,232 @@
...
+         
+         It uses the components returned by the
+         components factory.
+         
+         - Returns: A valid login controller configuration to use.
+     */
+    internal func createLoginControllerConfiguration() -> LoginControllerConfiguration {
+        let createLoginView: () -> LoginViewType = { [unowned self] in
+            return self._componentsFactory
+                .createLoginView(self._componentsFactory
```
guidomb: I would use intermediate instance variable instead of chainning this method invocations to improve readability.

Authentication/AuthenticationCoordinator.swift
```diff
@@ -0,0 +1,232 @@
...
+}
+
+
+
+extension AuthenticationCoordinator: LoginControllerTransitionDelegate {
+    
+    public final func toSignup(controller: LoginController) {
+        if _componentsFactory.initialScreen == .Login {
+            let signupController = createSignupController()
+            controller.navigationController!.pushViewController(signupController, animated: true)
```
guidomb: I would avoid using `!`. I prefer using `?`.

guidomb:
This looks way better than what we used to have before! Great work @danielaRiesgo!!! I've just took a look at the general structure and how the code got organized. I didn't check if every method was correct. I assumed that it works if the demo application works.

Authentication/AuthenticationCoordinator.swift
```diff
@@ -0,0 +1,232 @@
...
+}
+
+/**
+    Coordinator to start the application and regulate screen transitions.
+    Takes care of the authentication process until the Main View Controller
+    of your app is shown. (The authentication process is only triggered
+    when necessary).
+    The authentication process includes login and signup logic.
+*/
+public class AuthenticationCoordinator<User: UserType, SessionService: SessionServiceType where SessionService.User == User> {
```
guidomb: Do we still need the `UserType`?

guidomb:
Regarding the `SessionServiceType` I think we can declare the variable as `SessionServiceType` instead of using generics or associated type (if we drop the `UserType`) protocol.

Authentication/AuthenticationComponentsFactory.swift
danielaRiesgo: I wrote in known issue something to discuss.

We should discuss if the session service should go in the ComponentsFactory or not. The issue with having it there is that it will need an associated type because SessionServiceType uses an associated type.

Authentication/AuthenticationComponentsFactory.swift
danielaRiesgo: I wanted it to be a convenience init but xcode told me to erase "convenience".
Nevertheless, do you think this init is useful? Because the user will most probably (not to say always) change the color and font palettes, and so create their own view configurations.

guidomb: I think is useful for the first days of a project when you want to start hacking and don't care about the design.

Authentication/AuthenticationCoordinator.swift
danielaRiesgo: In this case, we expect there to e a navigationcontroller because if not the app won't work as expected. So I think that using `!` is a good way avoid the situation of not working and not knowing why.

guidomb: yeah I guess that is a valid point.

woloxci:

guidomb:
👍

FranDepascuali:
👍

woloxci:

